PR #15, opened 2017-02-15
Feature/us6488 pin details

JoeKer1:


src/app/services/store.service.ts
```diff
@@ -39,6 +39,17 @@ export class StoreService {
     );
   }
 
+  public loadEmail(): any {
+    this.api.getAuthentication().subscribe((info) => {
+      debugger;
```
ilpadre: Ahem.

JoeKer1: what? 🗡

PR #31, opened 2017-02-28
US6286 - Search Bar UI

plachmann:


src/app/components/neighbors/neighbors.component.html
```diff
@@ -0,0 +1,8 @@
+<app-search-bar (viewMap)="viewChanged($event)" (search)="doSearch($event)"> </app-search-bar>
+<br>
+<div *ngIf=mapViewActive>
```
vchtelmakh: "He wrapped himself in quotations - as a beggar would enfold himself in the purple of Emperors." - Rudyard Kipling

PR #42, opened 2017-03-03
Feature/us6245 say hi email

ilpadre:


src/app/components/member-said-hi/member-said-hi.component.html
```diff
@@ -0,0 +1,3 @@
+<div>Success!</div>
+
+<input type="button"  class="btn btn-secondary btn-block" (click)="btnClick();" value="Return to map">
```
JoeKer1: Not sure what the best practice is, but i'm just adding this comment so you know. You can link directly to a route from html in angular 2 with RouterLink. https://angular.io/docs/ts/latest/guide/router.html

This totally works though

src/app/services/pin.service.spec.ts
```diff
@@ -0,0 +1,57 @@
+/* tslint:disable:no-unused-variable */
+
```
JoeKer1: Oh crap, I thought I had this in there :(

src/app/components/pin-details/pin-details.component.ts
```diff
@@ -50,12 +52,12 @@ export class PinDetailsComponent implements OnInit {
     //We are just appending a bunch of properties to our pin class, not necessarily composing a
     //new pin via a constructor.  This should be rectified.
     this.pin = this.route.snapshot.data['pin'];
+    this.user = this.route.snapshot.data['user'];
```
JoeKer1: This seems like an expensive way to get the first and last name, but it doesn't appear that the login service returns the last name, so I don't have any better ideas.

PR #72, opened 2017-03-15
US7193: Host accepts or denies request to join

JoeKer1:


src/app/components/pin-details/gathering/gathering-requests/gathering-requests.component.ts
```diff
@@ -21,22 +23,55 @@ export class GatheringRequestsComponent implements OnInit {
...
-                                   this.state.setLoading(false);
-                                },
-                   err => console.log(err));
+      .subscribe(inquiryList => {
+        this.inquiries = inquiryList.filter((inquiry) => {
+          return inquiry.placed == null;
+        })
+        this.state.setLoading(false);
+      },
+      err => console.log(err));
```
Doug-Shannon: We will have to come back here and implement a real error state.  Also, setLoading will never be set to false if it fails so it will spin forever.

src/app/components/pin-details/gathering/gathering-requests/gathering-requests.component.ts
```diff
@@ -21,22 +23,55 @@ export class GatheringRequestsComponent implements OnInit {
...
+        // TODO add profile pictures to template text (otherwise would've used a content block)
+        let templateText;
+        let bpd;
+        if (approve) {
+          templateText = '<div class="container"><div class="row text-center"><h3>Request accepted</h3></div>';
+        } else {
+          templateText = '<div class="container"><div class="row text-center"><h3>Request Denied</h3></div>';
+        }
+        // tslint:disable-next-line:max-line-length
+        templateText += `<br/><div class="row text-center"<span>${inquiry.firstName} ${inquiry.lastName.slice(0, 1)} has been notified</span></div></div>`;
```
Doug-Shannon: I think you are missing the period after the first initial of the last name.

src/app/models/group.ts
```diff
@@ -31,9 +31,10 @@ export class Group {
...
+   constructor($groupId: number, $groupName: string, $groupDescription: string, $groupTypeName: string, $ministryId: number, $congregationId: number, $congregationName: string, $primaryContactId: number, $primaryContactEmail: number, $startDate: string, $endDate: string, $reasonEndedId: number, $availableOnline: boolean, $remainingCapacity: number, $groupFullInd: boolean, $waitListInd: boolean, $waitListGroupId: number, $childCareInd: boolean, $meetingDayId: number, $meetingDay: string, $meetingTime: string, $meetingFrequency: string, $meetingFrequencyId: number, $address: Address, $targetSize: number, $kidsWelcome: boolean, $proximity: number, $Participants: Participant[], $groupTypeId) {
```
Doug-Shannon: should we make groupTypeID nullable so it doesn't break anything?

PR #82, opened 2017-03-16
Feature/us6804 host invites person

Doug-Shannon:


src/app/components/pin-details/gathering/invite-someone/invite-someone.component.ts
```diff
@@ -0,0 +1,65 @@
...
+                        `<p class="text text-center">${someone.firstname.slice(0, 1).toUpperCase()}${someone.firstname.slice(1).toLowerCase()} ${someone.lastname.slice(0, 1).toUpperCase()}. has been notified.</p>`,
+                        `pin-details/${this.participantId}`,
+                        BlandPageType.Text,
+                        BlandPageCause.Success
+                    );
+
+                    this.blandPageService.setBlandPageDetailsAndGo(bpd);
+                },
+                failure => {
+                    this.state.setLoading(false);
```
JoeKer1: will a growl go here?

src/app/models/user.ts
```diff
@@ -1,12 +1,14 @@
+import { Person } from './Person';
 
-export class User {
+export class User implements Person {
 
     firstname: string;
```
JoeKer1: these need to stay here if theyre in person.ts?

src/app/services/bland-page.service.ts
```diff
@@ -33,6 +33,10 @@ export class BlandPageService {
         this.go();
     }
 
+    public primed() {
+        return (this.blandPageDetails !== null && this.blandPageDetails !== undefined);
```
JoeKer1: i think this.blandPageDetails != null does both

src/app/services/pin.service.ts
```diff
@@ -93,4 +95,8 @@ export class PinService {
   public requestToJoinGathering(gatheringId: number): Observable<boolean> {
     return this.session.post(`${this.baseUrl}api/v1.0.0/finder/pin/gatheringjoinrequest`, gatheringId)
   }
-}
+
+  public inviteToGathering(gatheringId: number, someone: Person): Observable<boolean> {
```
JoeKer1: should this go in group service? we'll need to use it for group tool

src/app/validators/email-address.validator.ts
```diff
@@ -0,0 +1,9 @@
+import { AbstractControl, Validators } from '@angular/forms';
```
JoeKer1: 👍

src/app/components/pin-details/gathering/invite-someone/invite-someone.component.ts
Doug-Shannon: Yes, once that is all sorted out.  we will have to make a pass at a bunch of stuff

src/app/models/user.ts
Doug-Shannon: I wanted to make User inherit from Person because they share the same fields except password. Just makes sense.  Then you could pass a person in for a user.

src/app/services/pin.service.ts
Doug-Shannon: its kinda specific to gathering, the backend explicitly sets Gathering stuff

src/app/models/user.ts
Doug-Shannon: meant user in for a person

src/app/services/bland-page.service.ts
Doug-Shannon: good to know.  I thought I heard a while back that they wanted us using the triple equals for type equality

JoeKer1: Yeah I think most the time, we could ask The Island. I just figured less code was better but maybe it's more clear

src/app/services/pin.service.ts
JoeKer1: cool! i didn't even see the url just assumed

PR #106, opened 2017-03-24
Feature/us7414 map view search optimization

eymiha:
Figure out the right zoom over a location (the highest level at least 10 pins), and preserve zoom when location changes and the new search-again-here button is clicked.


src/app/components/map/map.component.ts
```diff
@@ -34,19 +34,33 @@ export class MapComponent implements OnInit {
...
-          this.mapSettings.zoom = 15;
-          this.mapSettings.lat = pos.lat;
-          this.mapSettings.lng = pos.lng;
-        }
-      );
-    } else {
-      this.mapSettings.zoom = 15;
-      this.mapSettings.lat = this.searchResults.centerLocation.lat;
-      this.mapSettings.lng = this.searchResults.centerLocation.lng;
+    ////
```
plachmann: Remove commented code

src/app/components/search-local/search-local.component.ts
```diff
@@ -0,0 +1,41 @@
...
+      if ((update.value === "dragend") || (update.value === "zoom_changed")) {
+        this.mapView = update;
+        this.showButton();
+      }
+    });
+  }
+
+  public active: boolean;
+
+  ngOnInit() {
```
plachmann: We need some unit tests for this new component

src/app/services/state.service.ts
```diff
@@ -26,4 +27,12 @@ export class StateService {
   public getShowingPinCount() {
     return this.showingPinCount;
   }
+
```
plachmann: add some unit tests for these new methods

src/app/components/map/map.component.ts
eymiha: commented code block removed.

src/app/components/search-local/search-local.component.ts
eymiha: tests added

src/app/services/state.service.ts
eymiha: tests added

PR #108, opened 2017-03-27
Feature/us7424 frontent caching

Doug-Shannon:


config/webpack.common.js
```diff
@@ -85,7 +85,7 @@ module.exports = {
         to: 'fonts/'
```
sseissiger: Just putting in comment so we can block MERGE!

Doug-Shannon: good call

src/app/components/map/map.component.ts
```diff
@@ -38,14 +36,14 @@ export class MapComponent implements OnInit {
...
-      this.userLocationService.GetUserLocation().subscribe(
-        pos => {
-          this.mapSettings.zoom = 15;
-          this.mapSettings.lat = pos.lat;
-          this.mapSettings.lng = pos.lng;
-        }
-      );
+      // this.state.setLoading(true);
+      // this.userLocationService.GetUserLocation().subscribe(
+      //   pos => {
```
JoeKer1: should this stuff be removed? May need to ask analog

src/app/models/group.ts
```diff
@@ -74,5 +74,7 @@ export class Group {
         this.participantCount = $participantCount;
     }
 
-
+    public static overload_Constructor_One(groupId: number, participants: Participant[]): Group {
```
JoeKer1: They should make it easier to do this

src/app/services/base-service/cache.service.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        expect(partial).toBe(false);
+        expect(full).toBe(true);
+    }));
+    
+    it('should clearCache', inject([CacheableService], (service: CacheableService<number>) => {
+        service['cache'] = 123;
+        service['cacheLevel'] = CacheLevel.Full;
+        service['userIdentifier'] = 3
+
+        service['clearCache']();
```
JoeKer1: Does service.clearCache(); not work? Pretend this comment is on all the method calls. 

Setting a property on the service like that is fine. Oh is it private?

src/app/services/base-service/smart-cache.service.spec.ts
```diff
@@ -0,0 +1,569 @@
+import { TestBed, async, inject } from '@angular/core/testing';
+import { SmartCacheableService, CacheLevel } from './cacheable.service';
+
+describe('Service: SmartCachableService', () => {
```
JoeKer1: The amount and level of testing in here is great. 🥇 You can replace John at the top of mount TDD

src/app/services/session.service.ts
```diff
@@ -110,27 +125,119 @@ export class SessionService {
...
-  public getRequestOption(options?: RequestOptions):  RequestOptions {
+  public getRequestOption(options?: RequestOptions): RequestOptions {
     let reqOptions = options || new RequestOptions();
     reqOptions.headers = this.createAuthorizationHeader(reqOptions.headers);
     return reqOptions;
   }
 
   public getContactId(): number {
-    return +this.cookieService.get(this.contactId);
+    let cID = +this.cookieService.get(this.contactId);
```
JoeKer1: good call, thanks for fixing.

src/app/services/session.service.ts
```diff
@@ -110,27 +125,119 @@ export class SessionService {
...
+    return this.get(this.baseUrl + 'api/v1.0.0/authenticated')
+      .map((res: Response) => {
+        return res || this.defaults.authorized;
+      })
+      .catch((res: Response) => {
+        return [this.defaults.authorized];
+      });
+  }
+
+  public isLoggedIn(): boolean {
```
JoeKer1: I think we should just have isLoggedIn() and logOut() but I can take care of that later

JoeKer1:
I think a readme.md file in the services folder that goes over the two caching types and how they work with examples would be super helpful. Not necessary for this PR but something someone should do at some point. (that someone doesn't have to be you)

JoeKer1:
Nice job, especially with unit tests. This was a huge change.

src/app/models/group.ts
```diff
@@ -35,13 +35,13 @@ export class Group {
...
-    $meetingFrequencyId: number, $address: Address, $targetSize: number, $kidsWelcome: boolean, $proximity: number,
-    $Participants: Participant[], $groupTypeId: number = null, $participantCount: number) {
+    constructor($groupId?: number, $groupName?: string, $groupDescription?: string, $groupTypeName?: string, $ministryId?: number,
+    $congregationId?: number, $congregationName?: string, $primaryContactId?: number,
+    $primaryContactEmail?: number, $startDate?: string, $endDate?: string, $reasonEndedId?: number, $availableOnline?: boolean,
+    $remainingCapacity?: number, $groupFullInd?: boolean, $waitListInd?: boolean, $waitListGroupId?: number,
+    $childCareInd?: boolean, $meetingDayId?: number, $meetingDay?: string, $meetingTime?: string, $meetingFrequency?: string,
+    $meetingFrequencyId?: number, $address?: Address, $targetSize?: number, $kidsWelcome?: boolean, $proximity?: number,
+    $Participants?: Participant[], $groupTypeId?: number, $participantCount?: number) {
         this.groupId = $groupId;
```
plachmann: Why are we making all of these nullable?

src/app/models/group.ts
Doug-Shannon: agreed.  We need a standard though.

src/app/models/group.ts
Doug-Shannon: Sometimes you need a group with just a group ID and participants and instead of having to build a full group object when you don't need it, or creating a new DTO, you can construct a partial group.  I added one overloaded constructor, but this pattern would allow us to create as many as we needed.

PR #111, opened 2017-03-27
Feature/us7227 add labels

vchtelmakh:
---


src/app/components/canvas-map-overlay/canvas-map-overlay.component.css
```diff
@@ -0,0 +1,7 @@
+canvas {
+    position: absolute;
```
plachmann: Is absolute positioning OK for microclient?

vchtelmakh: Absolute is ok, fixed is not.

PR #124, opened 2017-03-28
US6921 - App Header

seancdavis:
As a Crossroads Connect user, 
I want to see the Connect branding in the top header of the Connect application


seancdavis:
@tcmacdonald: Still need tests here (and still have that issue with the app test failing). I also should point out that I stripped the search bar out of the header and put it back into the neighbors component because it was causing too much of an issue.

seancdavis:
@tcmacdonald: Test suite issue fixed, so now we're good minus writing new tests for features here.

tcmacdonald:
Sorry @seancdavis – build failed. Can you take another look?

seancdavis:
@tcmacdonald: Just pushed a couple more changes to help with the labels on the map falling behind the header (the labels were not respecting their boundaries for some reason).

seancdavis:
@tcmacdonald: I think I did this correctly. @zarahzachz, @boggsey, and were overlapping a lot today, but I think I had this is ready to go. I had to move the localized search button.

PR #128, opened 2017-03-29
Feature/us7166 return to last spot on map

Doug-Shannon:
Small PR to enable a return to the map from where you used to be.

### DIFF
--- a/src/app/components/map/map.component.html
+++ b/src/app/components/map/map.component.html
@@ -3,8 +3,8 @@
 <canvas-map-overlay></canvas-map-overlay>
 
 <sebm-google-map id="connect"
-                 [latitude]="searchResults ? searchResults.centerLocation.lat : mapSettings.lat"
-                 [longitude]="searchResults ? searchResults.centerLocation.lng : mapSettings.lng"
+                 [latitude]="mapSettings.lat"
+                 [longitude]="mapSettings.lng"
                  [zoom]="mapSettings.zoom"
                  [disableDefaultUI]="mapSettings.disableDefaultUI"
                  [zoomControl]="mapSettings.zoomControl"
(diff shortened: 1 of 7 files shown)


Doug-Shannon:
It already works that way.  If you were on the map and go back it returns you to list view.

src/app/components/map/map.component.html
```diff
@@ -3,8 +3,8 @@
 <canvas-map-overlay></canvas-map-overlay>
 
 <sebm-google-map id="connect"
-                 [latitude]="searchResults ? searchResults.centerLocation.lat : mapSettings.lat"
-                 [longitude]="searchResults ? searchResults.centerLocation.lng : mapSettings.lng"
+                 [latitude]="mapSettings.lat"
```
vchtelmakh: I am uncomfortable with the searchResults ternary being removed. It's convoluted, but there was a reason we left it in there - I can't tell you what right now, but I remember testing heavily and something breaking when I tried to simplify this code.

sseissiger: Doug -  I'm trying to understand the problem you are solving and the current behavior (before your change).  Before your change were you getting the correct search results, just not centered in the correct position?
If so, you should be using searchResults.centerLocation.lat and searchResults.centerLocation.lng

Doug-Shannon: @sseissiger 
So our change, is to enable the user to be returned to the exact spot on the map they were at, lat/long/zoom when they clicked on a pin.

For instance, I load the page and see Oakley, I then move the map super far North and click on one of my pins in Trenton.  I return to the map and see the same view that I saw when I left the map (Trenton, not Oakley).

@vchtelmakh 
I think the reason it didn't work before was because the weirdness between the neighbors component and the map component both making calls to find the user location was causing multiple return sets at different times.  Now that that is gone, I think it simplifies it quite a bit.  OnInit is setting MapSettings to the SearchResults centered location anyways, so it is pretty redundant.  

If you still feel uncomfortable with it, do you have a better idea of how to implement this?  Right now, search results will always override what is in mapSettings.

PR #149, opened 2017-03-31
Defect/de3098 full address

Doug-Shannon:


src/app/models/pin.ts
```diff
@@ -36,4 +36,8 @@ export class Pin {
         this.pinType = pinType;
         this.proximity = proximity;
     }
+
+    public static overload_Constructor_One(){
```
JoeKer1: space before braces 

snitches get stitches

PR #154, opened 2017-03-31
Defect/de3166 show site addresses

vchtelmakh:
---


vchtelmakh:
@eymiha You are absolutely right - this was a quick patch.

PR #159, opened 2017-04-04
Feature/us7473 bland page layout

zarahzachz:
I've updated the styles/markup for all of the bland pages to match the approved design and modified the existing fauxdals to match the updated markup/styles.

Note: The transactional bland pages should have stacked avatars indicating a connection between different users but this feature appears to not be developed just yet. A new story will be added to Ample's backlog to capture this.

Corresponds with crdschurch/SS-CMS#284

----------
https://invis.io/VRAO3U5ST#/221711951_V2-02-03-02b-Finder-Connect-Results-Unlisted-Add desktop approved confirmation screen look & feel


zarahzachz:
@tcmacdonald The page has a content-block in it and it looks like it's not loading correctly in this instance but I can't recreate this behaviour.

This is what I get when I remove the content block entirely:
screen shot 2017-04-05 at 11 25 13 am

zarahzachz:
@tcmacdonald I added a `min-width` to the address block in addition to the updated content block IDs.

zarahzachz:
@tcmacdonald I fixed the media query widths for the address form. Also added a new class to vertically center the confirmation/transactional bland pages. This class won't work with the Add Me to Map page because the content there is too long to center vertically.

PR #161, opened 2017-04-04
Defect/de3148 login component width

zarahzachz:
Fix layout on signin screen and bugs that trickled into other layouts (getting started and map-list).

Corresponds with crdschurch/SS-CMS#285

----------
Login Page on Connect is too wide, it fills up the full screen.


zarahzachz:
@tcmacdonald I forgot about the registration page! Added the correct wrapper class to it so it'll have the same layout as signin.

Ready to review.

PR #164, opened 2017-04-05
DE3154 iOS Icons

tcmacdonald:


src/app/layout/header/header.component.html
```diff
@@ -3,9 +3,21 @@
   <!-- Brand Bar -->
   <div class="brand-bar" *ngIf="state.hasBrandBar">
     <div class="app-logo">
-      <svg viewBox="0 0 75 15">
-        <use xlink:href="/assets/brandbanner-logo-connect.svg#brandbanner-logo-connect"></use>
+
+      <svg width="75px" height="15px" viewBox="0 0 75 15" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+        <g id="Page-1" stroke="none" stroke-width="1" fill="#FFFFFF" fill-rule="evenodd">
```
seancdavis: Curious why you're writing the entire SVG inline here?

The only issue with this is that because you're not using the `<use>` approach, you are setting the `fill` directly. In this case, it's NBD because you know it's not going to change color.

tcmacdonald: Good question... I'm doing this because of some issues related to symlinking assets into the Phoenix application across multiple environments. If we need to revisit this approach later because they want a different color, we can do that.

PR #175, opened 2017-04-06
Defect/justlinting

JoeKer1:


JoeKer1:
This PR gets rid of most of the silly errors we have. The rest will need more thought (== vs ===)

PR #181, opened 2017-04-09
Defect/de3216 back button not working

JoeKer1:


src/app/components/authentication/authentication.component.spec.ts
```diff
@@ -61,7 +64,6 @@ describe('Component: Authentication', () => {
   });
 
   it('should call the router to move to the previous step', inject([LoginRedirectService], (loginRedirectService) => {
```
JoeKer1: Real talk: I'm not sure why this is passing. Maybe because of the injected RedirectService? I don't know maybe we should delete it because I think this is a duplicate test of 'should call cancelRedirect when the back button is clicked'. Which I just realized it a much too long of a time (much how this sentence has just run on and on and on). Should probably rename this 'back() should call cancelRedirect()'

src/app/services/login-redirect.service.ts
```diff
@@ -35,12 +38,12 @@ export class LoginRedirectService {
...
 
-  public cancelRedirect(target = this.DefaultAuthenticatedRoute ): void {
-    if (this.originalTarget) {
+  public cancelRedirect(): void {
+    if (this.origin) {
       this.redirectFunction = null;
-      this.router.navigate([this.originalTarget]);
+      this.router.navigateByUrl(this.origin);
     } else {
-      this.router.navigate([target]);
```
JoeKer1: not sure why but target was always undefined anyways.

JoeKer1: So what this should now do is navigate back to the route from whence it came by path instead of by state. when redirectToLogin() is called it stored the path in the origin variable. I also noticed it _doesn't_ seem to do anything with params which we may need to fix when group tool is assimilated.

src/app/components/host-application/host-application.component.ts
```diff
@@ -29,9 +29,6 @@ export class HostApplicationComponent implements OnInit {
   }
 
   public ngOnInit() {
-    if (!this.session.isLoggedIn()) {
```
JoeKer1: We should notify the team to _not_ do this and use the route guard instead. I think this was old and before the route guard existed anyways, but just in cases.

src/app/components/pin-details/gathering/gathering.component.ts
```diff
@@ -47,7 +47,7 @@ export class GatheringComponent implements OnInit {
     private content: ContentService) { }
 
   public ngOnInit() {
-    window.scrollTo(0,0);
+    window.scrollTo(0, 0);
```
JoeKer1: I know why this was done, I haven't figured out why it was scrolling half way down the page to begin with. But I would like to know the answer to the question to know if this is the best solution.

PR #185, opened 2017-04-10
DE3219 - No Results

seancdavis:
Add a title and bump content into the middle of the page

@zarahzachz: You can get to this at `/no-results`.


zarahzachz:
@seancdavis I'm having problems accessing this page. I get a spinner that spins for eternity and no console errors.

zarahzachz:
@seancdavis I finally got it working. There's some changes I would recommend, but I don't know if it's part of the scope of this defect. Mainly, I think the header is wrong in the content block. Should be either and `h1` or `h2` and the font style should be in-line with the other header font styles on the site, i.e., that `<strong>` tag needs to go (and maybe add a `.title` class?).

seancdavis:
@zarahzachz: I received input from Rob and adjusted so we could make sure the header is centered. Meanwhile US7830 will fix the buttons.

I added the native header to be consistent with other pages. Still waiting on Rob's feedback as to whether he wants this removed or not.

zarahzachz:
@seancdavis Rob's design had the text centered on this page, but it's coming up as left aligned for me.

screen shot 2017-04-13 at 9 16 28 am

seancdavis:
@zarahzachz: I merged with development and brought it up to date. This has been centered for me since yesterday, though (which is the CSS I added). Not sure why you wouldn't see that. Try again and we may have to look together.

zarahzachz:
@seancdavis I just did a hard refresh and pulled the latest. I'm still not seeing centered text.

PR #193, opened 2017-04-10
Feature/us7672 unique site label

vchtelmakh:
---


src/app/models/pin-label-data.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.lastInitial = lastInitial;
+    this.siteName = siteName;
+    this.isHost = isHost;
+    this.isMe = isMe;
+    this.pinType = pinType;
+  }
+
+}
+
+export class PinLabel {
```
plachmann: There is some logic in this model that is probably worthy of a unit test

src/app/services/google-map.service.ts
```diff
@@ -152,10 +153,23 @@ export class GoogleMapService {
...
-    let decDeg = parseInt('' + dec);
-    let decMin = parseInt('' + (dec - decDeg) * 60.0);
-    let decSec = parseInt('' + (dec - decDeg - (decMin / 60.0)) * 3600.0);
+    let decDeg = parseInt('' + dec, 10);
+    let decMin = parseInt('' + (dec - decDeg) * 60.0, 10);
+    let decSec = parseInt('' + (dec - decDeg - (decMin / 60.0)) * 3600.0, 10);
     return `${decSgn}${decDeg}° ${decMin}' ${decSec}"`;
   }
 
+  public getLabelHeightAdjustment(cHeight: number, marker: MapMarker, isHostOrMe: boolean): number {
```
plachmann: Add a unit test for this new method.

src/app/services/pin-label.service.ts
```diff
@@ -0,0 +1,87 @@
+import { Injectable, NgZone } from '@angular/core';
+
```
plachmann: Need unit tests for new service

src/app/services/google-map.service.ts
vchtelmakh: Done.

src/app/services/pin-label.service.ts
vchtelmakh: Done.

src/app/models/pin-label-data.ts
vchtelmakh: Done.

PR #202, opened 2017-04-11
US7830 - Vertical Button Groups

seancdavis:
Move vertical button groups to crds-styles

---

Connected to crdschurch/crds-styleguide#100 and crdschurch/crds-styles#92.


seancdavis:
@tcmacdonald: Great idea. I think I had the others accounted for (bland and add me ...). This last commit covers Getting Started. Did I miss anything else?

tcmacdonald:
@seancdavis I have no idea :neckbeard: maybe @zarahzachz can weigh in? She'll know!

src/app/components/address-form/address-form.component.html
```diff
@@ -85,7 +85,7 @@
     </div>
   </div>
 
-  <footer class="page-footer page-footer-vertical">
+  <footer class="page-footer page-footer-vertical btn-group-vertical">
```
zarahzachz: `page-footer-vertical` looks like an unused class. Should be removed from markup.

src/app/components/getting-started/getting-started.component.html
```diff
@@ -28,10 +28,15 @@ <h3 class="entry-header">{{pin.name}} <span *ngIf="pin.host" class="label label-
...
+    <div class="btn-group-vertical">
+      <a routerLink="/add-me-to-the-map" class="btn btn-secondary">Add me to map</a>
+    </div>
 
     <!-- Host Text/Buttons -->
     <crds-content-block id="finderGettingStartedHost" class="text-center"></crds-content-block>
-    <a routerLink="/host-signup" class="btn btn-secondary btn-block">Sign up</a>
-    <a routerLink="/whats-a-host" class="btn btn-secondary btn-block">What's a host</a>
+    <div class="btn-group-vertical">
+      <a routerLink="/host-signup" class="btn btn-secondary btn-block">Sign up</a>
```
zarahzachz: `.btn-block` is a class meant to give buttons full-width. Since these buttons are not meant to be full-width, it looks like the `.btn-group-vertical` is overriding `.btn-block`? I don't know if this is the behaviour we want, but it feels like there's some excess CSS here when removing a class would suffice.

src/app/components/no-results/no-results.component.html
```diff
@@ -1,17 +1,15 @@
...
-         (click)="btnClickBecomeHost();" value="Become a Host">
-
-  <input type="submit" id='btnFindOnlineGroup'
-         class="btn btn-secondary btn-block"
-         (click)="btnClickFindOnlineGroup();" value="Find an online group">
-
-  <input type="submit" id='btnBack' class="btn btn-secondary btn-block"
-         (click)="btnClickBack();" value="Try Again">
+  <div class="btn-group-vertical">
+    <button type="submit" id='btnAddToMap' class="btn btn-secondary btn-block"
```
zarahzachz: Again, review the use of `.btn-block` here.

src/app/components/getting-started/getting-started.component.html
zarahzachz: 👍 Reviewed the button width used here. It looks like the decision was made to override Bootstrap's `.btn-group-vertical` width (which was 100%). It's just a coincidence this class was added to the same selector that had `.btn-block`

PR #208, opened 2017-04-12
fixed address, added lodash.  Im a super star.

Doug-Shannon:
Victor is my spirit animal.

### DIFF
--- a/src/app/components/add-me-to-map/add-me-to-map.component.ts
+++ b/src/app/components/add-me-to-map/add-me-to-map.component.ts
@@ -1,22 +1,23 @@
 import { Component, OnInit, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { FormGroup, FormBuilder, FormControl, Validators } from '@angular/forms';
 import { Router, ActivatedRoute } from '@angular/router';
+import { ToastsManager } from 'ng2-toastr';
 
 import { BlandPageService } from '../../services/bland-page.service';
 import { StateService } from '../../services/state.service';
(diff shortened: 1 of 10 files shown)


src/app/components/add-me-to-map/add-me-to-map.component.ts
```diff
@@ -27,20 +28,40 @@ export class AddMeToMapComponent implements OnInit {
...
+      () => {
+        this.state.setLoading(false);
+        this.ready = true;
+      })
+    .subscribe(
+      success => {
+        this.userData.address = success;
+      },
+      error => {
+        this.toast.error(this.content.getContent('errorRetrievingFullAddress'));
```
vchtelmakh: Should this really be a toast, from a business logic perspective? Or just display empty fields?

src/app/components/pin-details/person/person.component.ts
```diff
@@ -28,6 +28,7 @@ export class PersonComponent implements OnInit {
   }
 
   ngOnInit() {
+    //what is this here for?
```
vchtelmakh: This is here because microclient saves scroll position, so we end up looking at the footer when we navigate to certain pages from the bottom of the list view.

JoeKer1: I believe there was an issue when these components loaded it would load and be scrolled to the bottom. I don't know what causes that but would like to know because i'm not sure this is the best solution

JoeKer1: 👍

PR #211, opened 2017-04-12
Defect/de3233 solve aws lag

sseissiger:


src/app/services/state.service.ts
```diff
@@ -79,4 +82,12 @@ export class StateService {
...
   }
 
+  public setPostedPin(pin: Pin) {
+    this.postedPin = pin;
+  }
+
+  public getPostedPin() {
+    return this.postedPin;
+  }
+
```
vchtelmakh: As far as I can tell, we don't need getters and setters for services anymore - just set the value directly. Could be missing some issue, but it does work to directly get/set var.

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -175,4 +177,22 @@ export class NeighborsComponent implements OnInit {
     this.router.navigateByUrl('/no-results');
   }
 
+  private foundPinElement = (pinFromResults: Pin): boolean => {
+    let postedPin = this.state.getPostedPin();
+    return (postedPin.participantId === pinFromResults.participantId && postedPin.pinType === pinFromResults.pinType);
+  }
+
+  private verifyPostedPinExistance() {
```
vchtelmakh: existence

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -84,7 +84,9 @@ export class NeighborsComponent implements OnInit {
     this.mapViewActive = isMapViewActive;
   }
 
-  processAndDisplaySearchResults(results: PinSearchResultsDto, searchString, lat, lng, zoom) {
+  processAndDisplaySearchResults(searchString, lat, lng): void {
+    // include posted pin if not included in results
+    this.verifyPostedPinExistance();
```
vchtelmakh: existence***

PR #212, opened 2017-04-12
DE3261 - Send in the proper viewtype

plachmann:


src/app/services/google-map.service.ts
```diff
@@ -82,6 +82,8 @@ export class GoogleMapService {
     let pop = this.countPopAtZoom(bounds, zoom, pins, pops);
     if (pop < popTarget) {
       if (zoom <= 3) {
+        console.log('pop=' + pop + ' popTarget=' + popTarget);
```
sseissiger: logging...

PR #224, opened 2017-04-13
DE3284 - Fauxdal Error Style

seancdavis:
Fix width on alert on add me to map screen.

The easiest way to have the alert popup is to go to `address-form.component.html` and change this:

```html
<div *ngIf="submissionError" class="alert alert-danger">
```

to this:

```html
<div class="alert alert-danger">
```


zarahzachz:
@seancdavis The text is center-aligned in that alert block. The DDK has alerts as left-aligned.

Bootstrap has a `.text-left` class, but it doesn't work because I've set the content-block as being center-aligned and this error message is being controlled by a content-block.

seancdavis:
Good call, @zarahzachz. I'll add a connect-specific style to control the alignment.

PR #236, opened 2017-04-17
DE3299 - Input shadow on iOS

zarahzachz:
Add local style to fix shadow on searchbar input field. Only visible in iOS.

*This style already exists in `crds-styles/development` but because we've locked the versions, `crds-connect` can't access this updated style. This style is added as a fix until it can access the latest from `crds-styles/development`*


zarahzachz:
@seancdavis I updated the `package.json` file and removed the redundant CSS. 

Ready to review.

PR #239, opened 2017-04-17
Don't redirect if we're already where we wanna be

JoeKer1:


Doug-Shannon:
You don't have to go home but you caaaaant staaaay heeeeeeeeere

PR #250, opened 2017-04-20
Feature/us7829 css refactor list view

zarahzachz:
Welcome to Refactor Land!
Updated the list view and pin details header to use Bootstrap's media object naming convention. 

Bootstrap defines a media object as: 
> Abstract object styles for building various types of components (like blog comments, Tweets, etc) that feature a left- or right-aligned image alongside textual content.

Corresponds with crdschurch/crds-styles#100

----------
Move the styles for the media objects from crds-connect to crds-styles. 
Media objects include the molecule that makes up the user details as shown on pin details & list view (includes avatar image, name, description, address, etc)

Part of a larger task to clean up connect.scss, clear out excess not in use and/or covered within crds-styles


src/styles/application.scss
```diff
@@ -126,6 +130,12 @@ app-add-me-to-map {
     padding-top: 3rem;
   }
 
+  .fauxdal-content {
```
seancdavis: Why part of this commit?

zarahzachz: Oh yeah, I saw that this style got wiped out in a previous merge (not sure where) so instead of creating a new story/defect - I added it to this story.

PR #264, opened 2017-04-25
Feature/us7426 refactor unit tests

ilpadre:


src/app/components/map-content/map-content.component.spec.ts
```diff
@@ -1,27 +1,38 @@
...
 describe('MapContentComponent', () => {
   let component: MapContentComponent;
   let fixture: ComponentFixture<MapContentComponent>;
+  let mockGoogleMapService,
+      mockGoogleMapsAPIWrapper,
+      mockMapsAPILoader,
+      mockStateService;
 
   beforeEach(async(() => {
+        mockGoogleMapService = jasmine.createSpyObj<GoogleMapService>('sessionService', ['constructor']);
```
JoeKer1: does 'constructor' here like actually call the constructor and create spies for all the methods in there? if not we should pass in the methods used here.

src/app/components/search-local/search-local.component.ts
```diff
@@ -17,27 +17,27 @@ export class SearchLocalComponent implements OnInit {
   public active: boolean;
 
   constructor(public mapHelper: GoogleMapService,
-              private state: StateService,
-              public search: SearchService,
-              private angulartics2: Angulartics2) {
-    mapHelper.mapViewUpdatedEmitter.subscribe((update) => {
```
JoeKer1: Did you test this locally after moving to ngOnInit()?

PR #265, opened 2017-04-25
DE3283 adding Angulartics2 to all components (capture pageviews from …

sseissiger:
…route)


JoeKer1:
I am not sure if importing angularitics does anything by itself.

PR #274, opened 2017-04-26
US7922 - Host Sign Up

seancdavis:
https://invis.io/Y6A640O35#/215891553_02-03-04b-Finder-Connect-Results-Listed-Learnmore

faudal + form page
hide/show on home/not home address

This should require fairly little work. This is a 3 step process:
1. What's a host (content block & btns)
2. Host info form (content block, form fields, validation, btns)
3. Confirmation page (content block, btns) -- unknown if this pg will use the blandPageService?

Most of these elements should already exist no problem in crds-styles and these pages should use the fauxdal template which is largely good to go. Just need to assemble all elements and clean up as needed.

### DIFF
--- a/src/app/components/host-application/host-application.component.html
+++ b/src/app/components/host-application/host-application.component.html
@@ -9,7 +9,9 @@
                 <!--HOME ADDRESS-->
                 <div class="control-label">HOME ADDRESS:</div>
                 <address-form [parentForm]="hostForm" [address]="homeAddress" groupName="homeAddress" [isFormSubmitted] = "isFormSubmitted"></address-form>
-                <div>Any changes you make to your home address will be reflected in your profile on crossroads.net</div>
+                <div class="text-muted font-size-smaller">
+                  Any changes you make to your home address will be reflected in your profile on crossroads.net
+                </div>
                 <!--MEETING AT HOME? -->
                 <div class="row">
                     <div class="form-group col-xs-12">
@@ -55,7 +57,6 @@
                 <div class="control-label push-half-bottom">
                   GATHERING DESCRIPTION:
                 </div>
-                <crds-content-block id="hintsForGatheringDesc"></crds-content-block>
 
                 <div class="row">
                     <div class="form-group col-xs-12">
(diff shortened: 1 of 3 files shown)


src/app/components/host-application/host-application.component.html
```diff
@@ -1,85 +1,85 @@
...
+                <div class="row">
+                    <div class="form-group col-xs-12">
+                        <label class="sr-only" for="gatheringDescription">GATHERING DESCRIPTION:</label>
+                        <textarea tabindex="1"
+                                  type="text"
+                                  class="form-control"
+                                  rows = "5"
+                                  maxlength="75"
+                                  placeholder="Group Description"
+                                  id = "gatheringDescription"
```
zarahzachz: The content block affiliated with this id is injecting some unescaped HTML into the `<textarea>` tag. I tried manually cleaning this up on the CMS, but it's being difficult.

src/app/components/host-application/host-application.component.html
```diff
@@ -1,85 +1,85 @@
...
+                                  formControlName="gatheringDescription"
+                                  name="gatheringDescription"></textarea>
+                        <p class="error help-block" *ngIf="isFormSubmitted && !hostForm.controls['gatheringDescription'].valid">Gathering description is invalid!</p>
                     </div>
-                </form>
-            </div>
+                </div>
+                <footer class="text-center">
```
zarahzachz: May want to bring in the `.btn-group-vertical` class so it matches the layout on Add Me to Map, or make the buttons on Add Me to Map match host-signup. Just add some consistency.

src/app/components/host-application/host-application.component.html
```diff
@@ -1,85 +1,85 @@
...
-                    </footer>
-                    <div *ngIf="submissionError" class="alert alert-danger">
-                        <crds-content-block id="generalError"></crds-content-block>
+                </div>
+                <br>
+                <!--GATHERING DESCRIPTION-->
+                <div class="control-label push-half-bottom">
+                  GATHERING DESCRIPTION:
+                </div>
+                <crds-content-block id="hintsForGatheringDesc"></crds-content-block>
```
zarahzachz: The CMS has some weird "bullet points" being used that need cleaning up.

PR #275, opened 2017-04-26
US7375 - Clean Up Confirmation Screens

seancdavis:
As a Connect user or host, I should see confirmation screens that match the approved design when I transact with other users.

---

All I did here was look for instances of `BlandPageType.Text` and fixed the markup so it would appear. In _many_ of these cases, there's really no easy way to make the confirmation page appear. And that's problematic for passing this on to QA, so I'm open to suggestions. (Perhaps this should sit here until all of the confirmation screens are working properly?)


zarahzachz:
I was able to access most of the sections updated. The only thing not working atm is sending emails (I got errors in every direction I tried).

PR #282, opened 2017-04-26
Defect/de3380 remove styles from content block

vchtelmakh:
DE3380 and DE3381 on one branch.


JoeKer1:
Should you use a pipe instead? it feels like we could be doing this in a lot of different places

PR #286, opened 2017-04-27
DE3389 - Gathering page layout

zarahzachz:
Fix layout so it's not ugly.

The header on gathering details page shouldn't have the group description or buttons of any kind in it.


zarahzachz:
@seancdavis fixed this defect in another story that got merged.

PR #288, opened 2017-04-27
Feature/us7915 text layout cleanup

zarahzachz:
- Text should be left-aligned everywhere but fauxdals (where they remain center aligned)
- Map buttons should be default size (remove `.btn-sm` class)
- "Update results" button has moved to the bottom center of the map screen
- The first letter of headers are capitalized 
- Using the gray background/box-shadow layout throughout the application (`/getting-started`, `/sign-in`, `registration`, etc.)

Corresponds with crds-styles/development

----------
On screens where the content is centered with a white background, change this so it matches the /list view (i.e., has light gray background with slight box-shadow around it)
On fauxdals where the content is vertically centered, move the content slightly up. The vertical centering creates an optical illusion on the content.
Increase spacing on between the form and submit button on Add me to Map screen
Fauxdals should stay entirely center aligned
Remove .btn-sm from map buttons
Update results buttons to be centered on map (consider lower placement on mobile?) https://invis.io/VRAO3U5ST#/230153224_V2c-My-Stuff-A


zarahzachz:
@tcmacdonald Found the regression after all. Fixed and ready to review.

PR #290, opened 2017-04-28
Defect/de3394 infinite spinner add map

sseissiger:


src/app/components/add-me-to-map/add-me-to-map.component.ts
```diff
@@ -64,7 +64,15 @@ export class AddMeToMapComponent implements OnInit, AfterViewInit {
...
-        this.toast.error(this.content.getContent('errorRetrievingFullAddress'));
+        let prepopulateError = new BlandPageDetails(
+            'Back To The Map',
+            'Unable to get details for the Add Me To The Map form, please try again later.',
+            BlandPageType.Text,
+            BlandPageCause.Error,
+            '',
+            ''
+          );
+          this.blandPageService.primeAndGo(prepopulateError);
```
JoeKer1: I don't think we should show a bland page if we couldn't get addressline1 and addressline2...just make them fill it out again.

PR #292, opened 2017-04-28
DE3371 - Fauxdal Height

seancdavis:
There was an issue being able to scroll when the fauxdal was active, but only in iOS. This fix _should_ take care of it. Should be tested in the simulator while running maestro.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "bootstrap-sass": "3.3.7",
     "codelyzer": "2.0.0",
     "copy-webpack-plugin": "4.0.0",
-    "crds-ng2-content-block": "crdschurch/crds-ng2-content-block#development",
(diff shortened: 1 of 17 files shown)


zarahzachz:
@seancdavis Something's not quite right on desktop now... it might just be my computer. 

screen shot 2017-04-28 at 11 26 41 am

Also, the positioning of the text in iOS is placed a little high on the screen. I don't know if these are separate or related issues.
screen shot 2017-04-28 at 11 28 10 am

zarahzachz:
@seancdavis The host signup form isn't scrollable 😭

seancdavis:
@zarahzachz: I fixed the conflict. It _may_ have created a regression—it was a gnarly conflict. But I think it was mostly to do with indentation.

PR #315, opened 2017-05-02
Defect/de3439 host pin me pin

plachmann:


src/app/components/map-content/map-content.component.ts
```diff
@@ -189,8 +189,11 @@ export class MapContentComponent implements OnInit {
             let div = this.div_;
             div.className = 'pin-label';
             div.className += ' ' + pinType[this.labelData_.pinType].toLowerCase().toString();
-            if (this.labelData_.isMe) { div.className += ' me'; }
-            if (this.labelData_.isHost) { div.className += ' host'; }
+            if (this.labelData_.isMe) {
+               div.className += ' me';
+           } else {
```
vchtelmakh: Should this just be an `else if`?

PR #333, opened 2017-05-10
Feature/usip angular4

plachmann:


package.json
```diff
@@ -17,62 +18,67 @@
...
     "zone.js": "0.6.26"
   },
   "devDependencies": {
+    "@angular/cli": "1.0.1",
+    "@ngtools/webpack": "1.3.1",
     "@types/core-js": "0.9.34",
     "@types/jasmine": "2.5.41",
     "@types/node": "6.0.68",
-    "angular-cli": "1.0.0-beta.18",
+    "angular-router-loader": "^0.6.0",
```
vchtelmakh: It might be a good idea to lock the version on this just like we do on all the other packages.

src/app/components/pin-details/gathering/gathering.component.spec.ts
```diff
@@ -163,7 +167,7 @@ describe('GatheringComponent', () => {
         expect(mockToast.error).toHaveBeenCalledWith(expectedText);
     });
 
-    it('should redirectToLogin while request(ing)ToJoin', () => {
+    fit('should redirectToLogin while request(ing)ToJoin', () => {
```
vchtelmakh: Unfocus test, I think you mentioned you were working on this, just noting it so we don't forget.

PR #343, opened 2017-05-18
Feature/us8230 group search results

vchtelmakh:
https://rally1.rallydev.com/#/66097202244d/detail/userstory/116762444740

### DIFF
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -106,6 +106,7 @@ const appRoutes: Routes = [
   },
   { path: 'map', component: NeighborsComponent },
   { path: 'neighbors', component: NeighborsComponent },
+  { path: 'groupsv2', component: NeighborsComponent },
   { path: 'no-results', component: NoResultsComponent },
   {
     path: 'remove-person-pin/:participantId',
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -146,6 +146,13 @@ const appRoutes: Routes = [
       user: UserDataResolver
     }
   }, {
+    path: 'small-group/:groupId',
+    component: PinDetailsComponent,
+    resolve: {
+      pin: PinResolver,
+      user: UserDataResolver
+    }
+  }, {
     path: 'gathering/:groupId/edit',
     component: GatheringEditComponent,
     resolve: {
--- a/src/app/components/list-entry/list-entry.component.ts
+++ b/src/app/components/list-entry/list-entry.component.ts
@@ -42,7 +42,7 @@ export class ListEntryComponent {
     return this.type === pinType.GATHERING && this.contactId === this.currentContactId;
   }
 
-  public name() {
+  public formatName() {
     if(this.isSmallGroup()){
       return this.groupTitle ? this.groupTitle.toUpperCase() : '';
     } else {
--- a/src/app/components/list-entry/list-entry.component.ts
+++ b/src/app/components/list-entry/list-entry.component.ts
@@ -7,6 +7,8 @@ import { Address } from '../../models/address';
 import { SessionService } from '../../services/session.service';
 import { StateService } from '../../services/state.service';
 
+import { proximityUnavailableDefaultNum } from '../../shared/constants';
+
 @Component({
   selector: 'list-entry',
   templateUrl: 'list-entry.component.html'
@@ -108,7 +110,7 @@ export class ListEntryComponent {
   }
 
   public roundedProximity() {
-    return this.proximity ? this.proximity.toFixed(1) : 999;
+    return this.proximity ? this.proximity.toFixed(1) : proximityUnavailableDefaultNum;
   }
 
 }
--- a/src/app/components/list-entry/list-entry.component.ts
+++ b/src/app/components/list-entry/list-entry.component.ts
@@ -108,11 +108,7 @@ export class ListEntryComponent {
   }
 
   public roundedProximity() {
-    if(this.isSmallGroup()){
-      return 1;
-    } else {
-      return this.proximity.toFixed(1)
-    };
+    return this.proximity ? this.proximity.toFixed(1) : 999;
   }
 
 }
--- a/src/app/components/list-entry/list-entry.component.ts
+++ b/src/app/components/list-entry/list-entry.component.ts
@@ -17,6 +17,7 @@ export class ListEntryComponent {
   @Input() siteName: string = '';
   @Input() type: number;
   @Input() proximity: number = 0;
+  @Input() groupTitle: string = '';
   @Input() description: string = '';
   @Input() groupId: number = 0;
   @Input() address: Address = null;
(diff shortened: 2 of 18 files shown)


src/app/app.routing.ts
```diff
@@ -106,6 +106,7 @@ const appRoutes: Routes = [
   },
   { path: 'map', component: NeighborsComponent },
   { path: 'neighbors', component: NeighborsComponent },
+  { path: 'groupsv2', component: NeighborsComponent },
```
JoeKer1: It could just be groups right? groups is in crds-angular.

src/app/components/list-entry/list-entry.component.ts
```diff
@@ -17,6 +17,7 @@ export class ListEntryComponent {
   @Input() siteName: string = '';
   @Input() type: number;
   @Input() proximity: number = 0;
+  @Input() groupTitle: string = '';
```
JoeKer1: Could we replace all these inputs with just an input that is the entire pin object? Is that easier or make things worse?

src/app/components/list-entry/list-entry.component.ts
```diff
@@ -42,7 +43,11 @@ export class ListEntryComponent {
   }
 
   public name() {
```
JoeKer1: Should this be named formatName() now?

src/app/components/list-entry/list-entry.component.ts
```diff
@@ -99,7 +108,7 @@ export class ListEntryComponent {
   }
 
   public roundedProximity() {
-    return this.proximity.toFixed(1);
+    return this.proximity ? this.proximity.toFixed(1) : 999;
```
JoeKer1: What's the 999 for? Magic number? move to config?

src/app/components/list-view/list-view.component.ts
```diff
@@ -6,6 +6,8 @@ import { Pin } from '../../models/pin';
 import { PinSearchResultsDto } from '../../models/pin-search-results-dto';
 import { NeighborsHelperService } from  '../../services/neighbors-helper.service';
 
+import { App, app} from '../../shared/constants';
```
JoeKer1: not used?

src/app/components/map-content/map-content.component.ts
```diff
@@ -52,7 +52,9 @@ export class MapContentComponent implements OnInit {
 
   @HostListener('document:redrawingClusters', ['$event'])
   onClusterRedraw(event) {
-    this.drawLabels(event.data.markersNotInClusters);
+    if(this.state.activeApp === app.CONNECT){
```
JoeKer1: Maybe import add app and App in state.service and have this.state.isConnect?

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -55,9 +55,14 @@ export class NeighborsComponent implements OnInit, OnDestroy {
   }
 
   public ngOnInit(): void {
-    let haveResults = !!this.pinSearchResults;
 
-    if (!haveResults) {
+    this.state.setActiveApp(this.router.url);
+
+    let haveResults: boolean = !!this.pinSearchResults;
```
JoeKer1: Don't have to change this, but I think a lot of this stuff should move into the search service. All the data and logic could be moved there and maybe we could remove some emitters, maybe not.

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -99,45 +104,8 @@ export class NeighborsComponent implements OnInit, OnDestroy {
     this.verifyPostedPinExistence();
     this.ensureUpdatedPinAddressIsDisplayed();
 
-    // sort
     this.pinSearchResults.pinSearchResults =
```
JoeKer1: I still don't like pinSearchResults.pinSearchResults

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -99,45 +104,8 @@ export class NeighborsComponent implements OnInit, OnDestroy {
...
-              (p.firstName !== pl.firstName) ||
-              (p.lastName !== pl.lastName);
-            if (test) {
-              lastIndex = index;
-            }
-          return test;
-          }
-        }
-      );
+        this.pinService.sortPinsAndRemoveDuplicates(this.pinSearchResults.pinSearchResults);
```
JoeKer1: 👍

src/app/services/pin.service.ts
```diff
@@ -292,6 +294,65 @@ export class PinService extends SmartCacheableService<PinSearchResultsDto, Searc
     return pinSearchResults;
   }
 
+  public areLastResultsFromActiveApp (currentApp: string, lastResultsApp: string): boolean {
+    return currentApp === lastResultsApp;
+  };
+
+  public sortPinsAndRemoveDuplicates(pinSearchResults: Pin[]): Pin[] {
```
JoeKer1: 👍

src/app/app.routing.ts
vchtelmakh: We are calling it V2 because it's something that will replace crds-angular groups completely.

src/app/components/list-entry/list-entry.component.ts
vchtelmakh: 4sure y not

src/app/components/list-entry/list-entry.component.ts
vchtelmakh: Yes, magic number, will move to constants.

src/app/components/list-entry/list-entry.component.ts
vchtelmakh: I think it's easier and cleaner, but I was too lazy to do it when I noticed the same issue.

src/app/app.routing.ts
JoeKer1: It's a weird url I guess. Shouldn't it just be /connect/groups (at least right now)? We might need the urls to be crossroads.net/finder/connect and crossroads.net/finder/groups

src/app/components/list-view/list-view.component.ts
vchtelmakh: Word.

src/app/app.routing.ts
vchtelmakh: We have a story in the current sprint to change the url to something like crossroads.net/connect and crossroads.net/groups - no one wants to touch the story since its Maestro.

src/app/components/neighbors/neighbors.component.ts
vchtelmakh: Crimea

src/app/components/neighbors/neighbors.component.ts
vchtelmakh: Yes, having more than one line here is dirty. Not being sarcastic, logic in components == weaksauce.

src/app/components/map-content/map-content.component.ts
vchtelmakh: Not a bad idea.

JoeKer1:
Not sure why it didn't see the build so I went aheadand merged it.

PR #344, opened 2017-05-22
US8398 - Remove Map Fauxdal

seancdavis:
Make remove me from map page look like other fauxdals. You can get to this page from _your_ pin details, by clicking the "remove me from the map link" below the edit button.

---

Given a Connect user
When I remove myself from the map
Then I should see a success screen that utilizes the fauxdal and includes approved design/markup for fauxdal 

---

No corresponding PRs.

### DIFF
--- a/src/app/components/pin-details/person/remove-person-pin/remove-person-pin.component.html
+++ b/src/app/components/pin-details/person/remove-person-pin/remove-person-pin.component.html
@@ -1,22 +1,24 @@
-<div class="fauxdal fauxdal-fullscreen">
-  <div class="container-fluid">
-        <div class="pull-right close">
-                <a (click)="cancel()">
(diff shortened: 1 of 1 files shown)


zarahzachz:
@seancdavis The styles for the "remove me from map" fauxdal don't look like they match the style from other fauxdals. I could be wrong though...

screen shot 2017-05-22 at 4 11 09 pm

screen shot 2017-05-22 at 4 11 52 pm

The text looks off between these two. I'm not sure about the button placement either - because many of our buttons have become inline with each other and left-aligned. Might want to consult Rob on these.

seancdavis:
Good catch, @zarahzachz. This is _mostly_ from a content block. I made the changes in the CMS, and I added one line break here to make it more like the bland pages.

PR #345, opened 2017-05-22
DE3496 - remove the pin locally

plachmann:


src/app/services/pin.service.ts
```diff
@@ -168,7 +167,17 @@ export class PinService extends SmartCacheableService<PinSearchResultsDto, Searc
       return this.session.get(this.baseUrl + searchUrlZoom)
         // when we get the new results, set them to the cache
         .map(res => this.gatheringService.addAddressesToGatheringPins(res))
-        .do((res: PinSearchResultsDto) => super.setSmartCache(res, CacheLevel.Full, searchOptions, contactId))
+        .do((res: PinSearchResultsDto) => {
+          if ( this.state.removedSelf) {
+            this.state.removedSelf = false;
+            let index = res.pinSearchResults.findIndex(obj => obj.pinType === 1 && obj.contactId === contactId);
```
vchtelmakh: Would be better to bring in `import { Pin, pinType } from '../../../models/pin';` and then use the constant for the pin type: `obj.pinType === pinType.PERSON`.

plachmann: Yes it would. Good catch on the magic number

vchtelmakh:
Looks good to me!

PR #357, opened 2017-05-25
Add contact id, when available, to the call for pins. We now need it …

vchtelmakh:
DE3439: https://rally1.rallydev.com/#/66097202244d/detail/defect/113474173384


vchtelmakh:
Please do not merge until the API portion (https://rally1.rallydev.com/#/66097202244d/detail/defect/113474173384) is merged.

PR #361, opened 2017-05-30
Feature/us8250 participant can apply to be a leader

plachmann:


src/app/components/host-application/host-application.component.ts
```diff
@@ -46,10 +49,34 @@ export class HostApplicationComponent implements OnInit {
...
+    private appSettingsService: AppSettingsService,
+    private groupService: GroupService
   ) {}
 
   public ngOnInit() {
+    switch (this.appSettingsService.leadershipApplicationType) {
+      case LeadershipApplicationType.GROUP_LEADER:
+        this.groupService.getLeaderStatus().subscribe(
+          pos => {
+              console.log(pos);
```
sseissiger: Do you want to leave the logging in this file?  I'm fine leaving short term if you think it will help until the process is completely defined

PR #362, opened 2017-05-30
DE3528 - Remove me from map fauxdal

zarahzachz:
Add class of .fauxdal-open to the body to prevent duplicate scrollbar from popping up.
No corresponding PR.

**Check out fauxdals locally in Maestro to account for the length of the footer (this is what caused the duplicate scrollbar)**

---
Peter re-opened this defect. Spotted a duplicate scrollbar on the "Remove me from map" fauxdal when using Chrome on PC. Defect not visible on other fauxdals in the application.


zarahzachz:
@seancdavis I re-tested this in BrowserStack. Couldn't sign in on Edge, so the problem might still be there, but Chrome didn't show a secondary scrollbar for me. Can you re-check?

PR #363, opened 2017-05-31
DE3528 - Remove me from map fauxdal

zarahzachz:
Remove the click event from the 'Remove me from the map' link - replacing with a routerLink directive as this how other 'fauxdal triggers' have been wired. Removing the click event function from the TS file as it's no longer being used.

No corresponding PRs.

---
Peter noticed a brief flash of the double-scrollbars on the "Remove me from the map" fauxdal on all browsers. This behaviour isn't observed in any other fauxdal. Changing the click event to a dynamic `[routerLink]` seems to correct the issue and is consistent with how all other fauxdals are wired up.


zarahzachz:
@seancdavis I addressed the failing test problem. The tests weren't happy with my `[router-link]` change because it's monitoring for `this.state.setCurrentView` in each method. Unsure why cancel and "remove me from map" have different views affiliated/what purpose this serves but #YOLO!

PR #365, opened 2017-05-31
Recalculate proximity based on the map coords.

JoeKer1:


config/webpack.common.js
```diff
@@ -63,6 +63,13 @@ module.exports = {
       name: ['app', 'vendor', 'polyfills']
     }),
 
+    new webpack.ContextReplacementPlugin(
+         // The (\\|\/) piece accounts for path separators in *nix and Windows
+         /angular(\\|\/)core(\\|\/)@angular/,
+         helpers.root('./src'), // location of your src
+         { }
+       ),
```
vchtelmakh: u wut m8, y dis here

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -99,6 +100,13 @@ export class NeighborsComponent implements OnInit, OnDestroy {
 
   viewChanged(isMapViewActive: boolean) {
     this.mapViewActive = isMapViewActive;
+    if (!isMapViewActive) {
+          let location = this.state.getMapView();
+          let lastSearch = this.state.getLastSearch();
+          let coords = (location !== null ) ? location : lastSearch.coords;
```
vchtelmakh: Use types if they are known. I know it might just be "any" if we didn't create an object, in which case specifying it is kind of useless.

src/app/services/pin.service.spec.ts
```diff
@@ -14,13 +14,9 @@ import { Observable } from 'rxjs/Rx';
 import { Response, ResponseOptions } from '@angular/http';
 import { MockBackend } from '@angular/http/testing';
 
-import { Pin, pinType } from '../models/pin';
-import { User } from '../models/user';
-import { PinSearchResultsDto } from '../models/pin-search-results-dto';
+import { Pin, pinType, GeoCoordinates, User, PinSearchResultsDto, PinIdentifier, Address } from '../models';
```
vchtelmakh: =)

config/webpack.common.js
JoeKer1: To get rid of the compiler warnings

src/app/services/pin.service.ts
```diff
@@ -47,6 +49,7 @@ export class PinService extends SmartCacheableService<PinSearchResultsDto, Searc
   ) {
     super();
     this.SayHiTemplateId = sayHiTemplateId;
+    this.earthsRadiusInMiles = earthsRadiusInMiles;
```
vchtelmakh: Don't need this, can use imported value directly.

JoeKer1: sweet!

src/app/shared/constants.ts
```diff
@@ -3,6 +3,8 @@ export const crdsOakleyCoords: any = {
   lng: -84.423367
 };
 
+export const earthsRadiusInMiles: number = 3443.9;
```
vchtelmakh: Did you test after moving out? There is some weirdness with constants where they come up as undefined. Not sure that this applies here, but just as a heads up.

JoeKer1: Yup seems grood :D

PR #369, opened 2017-06-02
Feature/us7911 get out of my stuff view

vchtelmakh:


src/app/components/map-footer/map-footer.component.spec.ts
```diff
@@ -76,4 +88,10 @@ describe('Component: MapFooter', () => {
         expect(comp).toBeTruthy();
     });
 
+    it('should change state to my stuff if it is inactive and "My Stuff" is clicked', () => {
+        spyOn(comp, 'changeStateToMyStuff');
```
JoeKer1: Not sure this really tests anything other than that you called the function

vchtelmakh: It tests that the correct function was called - a different function is called if my stuff is active.

src/app/components/map-footer/map-footer.component.spec.ts
```diff
@@ -16,6 +16,16 @@ import { LoginRedirectService } from '../../services/login-redirect.service';
 import { UserLocationService } from  '../../services/user-location.service';
 import { AppSettingsService } from '../../services/app-settings.service';
 
+function fakenext(param: any) { return 1; }
```
JoeKer1: I might have a way this actually works, I haven't tested it yet though. It works in jest 🥇

src/app/components/search-bar/search-bar.component.html
```diff
@@ -2,10 +2,14 @@
...
       <input type="text"
-              class="form-control"
-              [(ngModel)]="searchText"
-              (keyup.enter)="onSearch(searchText)"
-              placeholder="Search zip or keyword...">
+             class="form-control"
+             [(ngModel)]="searchText"
+             (keyup.enter)="onSearch(searchText)"
+             placeholder="Search zip or keyword...">
+      <!--(keyup) = "searchKeyUp()"--> <!--This attribute belongs to the input above-->
```
JoeKer1: can this be removed?

src/app/services/state.service.ts
```diff
@@ -32,7 +32,7 @@ export class StateService {
   private myViewOrWorldView: string = 'world';
   private zoomToUse: number = -1;
   private savedMapView: MapView;
-  private lastSearch: SearchOptions;
+  public lastSearch: SearchOptions;
```
JoeKer1: Should this have a getter like the other things? I dunno probably doesn't matter.

src/app/components/search-bar/search-bar.component.html
vchtelmakh: I would rather not, Ample is going to use it soon.

src/app/components/map-footer/map-footer.component.spec.ts
vchtelmakh: Let me know.

PR #371, opened 2017-06-04
feature/US7911 Fix After CR

vchtelmakh:
---


vchtelmakh:
Ty for linty-minty.

vchtelmakh:
Roh, roh, will resolve after meeting.

PR #397, opened 2017-06-08
Defect/de3635 my stuff still inconsistent

vchtelmakh:


src/app/components/list-entry/list-entry.component.ts
```diff
@@ -16,6 +18,7 @@ import { proximityUnavailableDefaultNum, groupDescriptionLenth } from '../../sha
   templateUrl: 'list-entry.component.html'
 })
 export class ListEntryComponent {
+  @Input() pin: Pin;
   @Input() firstName: string = '';
```
JoeKer1: Does the pin have everything below inside of it? We can clean it up later if so

vchtelmakh: It most certainly does!

JoeKer1: 👍

PR #398, opened 2017-06-12
Defect/de3628 map grey zoom level

sseissiger:


package.json
```diff
@@ -35,7 +35,7 @@
     "angulartics2": "2.2.1",
     "copyfiles": "1.0.0",
     "core-js": "2.4.1",
-    "crds-styles": "0.5.0",
+    "crds-styles": "0.6.0",
```
plachmann: Follow up with Ample/Taylor on version.

PR #408, opened 2017-06-21
US8395 - add participant directly to group

plachmann:


AWS_Scripts/create_search_domain.ps1
```diff
@@ -2,7 +2,7 @@
 #
 # Before running you will need an AWS Access Key ID and an AWS Secret Access Key
 
-$searchdomain = "connect-dev"
+$searchdomain = "connect"
```
sseissiger: I am assuming this is correct - removing -dev, just double checking

PR #420, opened 2017-06-23
Defect/de3715 proximity gathering header

sseissiger:
I took roundedProximity function out of list-entry.component.ts and moved to service instead of duplicating the function in the gathering.component.ts

The readonly-address.component.ts line 27 is not using this service and is the cause of the last inconsistency I am seeing.  The address component is using 2 decimal places and the list view and gathering were using 1 decimal place.

I will be out tomorrow - feel free to refactor as the team sees fit.  Thanks.


vchtelmakh:
I don't think component-specific helpers are a bad thing. They add maintenance overhead, but they also do two positive things: Keep components clean and help avoid services with 500 lines of code. Instances of the latter are confusing even though "It's all in once place."

src/app/components/pin-details/gathering/gathering.component.ts
```diff
@@ -60,7 +60,7 @@ export class GatheringComponent implements OnInit {
     private angulartics2: Angulartics2,
```
JoeKer1: Can we get rid of address and proximity to display variables above?

src/app/components/pin-details/readonly-address/readonly-address.html
```diff
@@ -1,3 +1,3 @@
 <span *ngIf="showFullAddress"> {{showFullAddress ? address.addressLine1 : ""}} {{showFullAddress ? address.addressLine2 : ""}}<br></span>
 
-<span>{{address.city}}, {{address.state}} {{address.zip}} {{this.distString}}</span>
\ No newline at end of file
+<span>{{address.city}}, {{address.state}} {{address.zip}} {{distance |  number : '1.2-2' }}</span>
```
JoeKer1: Does this one need the MI on it?

sseissiger: It did not have it previously, so I left it off

src/app/components/pin-details/gathering/gathering.component.ts
sseissiger: I removed proximityToDisplay that we not used, but the address is used on line 88 to get the fullAddress (vs just city state zip)

src/app/components/pin-details/readonly-address/readonly-address.html
JoeKer1: It was in the component oninit

PR #421, opened 2017-06-23
DE3731 - Unique Contact

plachmann:


src/app/components/pin-details/gathering/add-someone/add-someone.component.ts
```diff
@@ -83,9 +80,8 @@ export class AddSomeoneComponent implements OnInit {
                 success => {
                     // display the modal so the user can choose 
                     this.state.setLoading(false);
-                    success.length > 0 ? this.matchFound = true : this.matchFound = false;
-                    this.matches = success;
-                    this.selectedMatch = this.matches[0];
+                    success ? this.matchFound = true : this.matchFound = false;
```
vchtelmakh: It's not immediately clear to me that success/obsevable 'next' is 'isMatchFound`. Maybe we should rename it, maybe not though, just a thought.

PR #422, opened 2017-06-23
Feature/us8266 leader can end participant

JoeKer1:


vchtelmakh:
"Feature/us8266 leader can end participant"

Added appropriate label for "end participant"

src/app/components/pin-details/gathering/participant-details/participant-details.component.html
```diff
@@ -1,22 +1,31 @@
-<div class="connect-container connect-layout-container is-gathering">
-<div class="row pin-details-dark dark-theme">
-    <div class="col-xs-12 soft media media-dark">
-        <profile-picture imageClass="media-object img-size-6" wrapperClass="media-img-wrapper" [contactId]="participant.contactId"></profile-picture>
-        <div class="media-body">
-            <h2 class="media-heading media-heading-stacked">
+<div class="connect-container connect-layout-container is-gathering" *ngIf="ready">
```
vchtelmakh: Should be 'readyEh' or 'isReady', may also want to specify what it is that is ready.

src/app/components/pin-details/gathering/participant-details/participant-details.component.ts
```diff
@@ -13,52 +13,60 @@ import { Component, Input, OnInit } from '@angular/core';
...
                         this.state.setPageHeader('Participant', this.redirectUrl);
                         this.state.setLoading(false);
-                    })
-                    .subscribe(
-                        (address) => {
-                            this.participantAddress = address;
-                        }, error => {
-                            console.log(error);
-                        });
+                    }).subscribe(address => {
```
vchtelmakh: You have a nested observable here, imo, the `next` event (upon receiving successful reply from getPartialPersonAddress) observable should be in a function. It's hard to immediately follow what's happening. This *is* nitpicky and you don't need to change it.

src/app/components/pin-details/gathering/participant-remove/participant-remove.component.ts
```diff
@@ -0,0 +1,74 @@
...
+            this.participantService.getGroupParticipant(this.groupId, this.groupParticipantId).finally(() => {
+                this.state.setLoading(false);
+            })
+            .subscribe(p => {}, error => {
+                console.log(error);
+                this.handleError();
+            });
+            this.redirectUrl = `/${this.router.url.split('/')[1]}/${this.groupId}`;
+        }, error => {
+            this.handleError();
```
vchtelmakh: Should loading be set to false here?

src/app/components/pin-details/gathering/participant-details/participant-details.component.html
JoeKer1: OH yeah I forgot about that, putting that in there gets rid of console errors. probably a better way to do it but it works for now

src/app/components/pin-details/gathering/participant-remove/participant-remove.component.ts
JoeKer1: it's in the finally block above

src/app/components/pin-details/gathering/participant-details/participant-details.component.ts
JoeKer1: Lets get together on Monday just so I'm sure of what you're saying :D

PR #432, opened 2017-06-27
US8009 - Mobile Search Keyboard

seancdavis:
Given a Connect user
when I use the search on the map/list of Connect
Then that input field should use the input type "search" so that the keyword can persist and has a clear button


zarahzachz:
@seancdavis Can you resolve merge conflicts please?

PR #437, opened 2017-06-29
Feature/us8261 create group

JoeKer1:


src/app/components/create-group/create-group-data.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Group } from '../../models';
+import { Address } from '../../models';
```
vchtelmakh: How do you get away without specifying full path, e.g.: '../../models/address';

Maybe `import {Address, Group } from '../..models' then?

src/app/components/create-group/create-group-summary/create-group-summary.component.ts
```diff
@@ -0,0 +1,16 @@
+import { StateService } from '../../../services/state.service';
+
+import { Component, OnInit } from '@angular/core';
```
vchtelmakh: Angular imports need to be imported before our services/components/models as per styleguide, not important.

src/app/components/create-group/page-1/create-group-page-1.component.ts
```diff
@@ -0,0 +1,35 @@
...
+            categories => {
+                this.categories = categories;
+                this.isComponentReady = true;
+            }
+        );
+    }
+
+    public onSelect(category: Category) {
+        category.selected = !category.selected;
+        console.log(category);
```
vchtelmakh: Remove console.log.

src/app/components/create-group/page-1/create-group-page-1.component.ts
```diff
@@ -0,0 +1,35 @@
...
+        this.state.setPageHeader('start a group', '/create-group');
+        this.initializeCategories();
+        this.state.setLoading(false);
+    }
+
+    private initializeCategories() {
+        this.lookup.getCategories().subscribe(
+            categories => {
+                this.categories = categories;
+                this.isComponentReady = true;
```
vchtelmakh: Add void return type. I think you can just use the spinner instead of component ready here.

src/app/models/category.ts
```diff
@@ -0,0 +1,24 @@
+import { Attribute } from './attribute';
+
+export class Category {
+    categoryId: number;
+    attribute: Attribute;
+    desc: string;
+    exampleText: string;
+    requiresActiveAttribute: boolean;
+    name: string;
+    selected: boolean = false;
```
vchtelmakh: I think setting values should be set in the constructor.

src/app/shared/MockTestData.ts
```diff
@@ -134,5 +129,24 @@ export class MockTestData {
         );
     }
 
+    public static getSomeCategories(numOfCategories: number = 5) {
```
vchtelmakh: Need to add return type. Five is magic, like friendship.

src/app/shared/MockTestData.ts
```diff
@@ -134,5 +129,24 @@ export class MockTestData {
...
+        let participants = new Array<Category>();
+        if (numOfCategories === 0) {
+            return null;
+        }
+        for (let index = 0; index < numOfCategories; index++) {
+            let participant = new Category(
```
vchtelmakh: Why is participant assigned to new cat?

vchtelmakh:
ParticipantService getAllLeaders test failed.

src/app/components/create-group/create-group-summary/create-group-summary.component.ts
JoeKer1: !important

src/app/shared/MockTestData.ts
JoeKer1: It's just for test otherwise I'd change it

src/app/shared/MockTestData.ts
JoeKer1: That copy pasta

PR #438, opened 2017-06-30
Feature/us8907 input icons buttons

zarahzachz:
Corresponds with:
crdschurch/crds-styles#162
crdschurch/crds-styleguide#175

---
Given a user on CR.net/connect
When I add text to the searchbar
Then I should see an 'X' icon/button to clear the text

Given a user on CR.net/groupsv2
When I see the searchbar
Then I should see a filter icon/button (equalizer)

When I add text to the searchbar
Then I should see an 'X' icon/button to clear the text

Given a developer/designer on DDK
When I view the forms > searchbar section
Then I should see a code example of a searchbar supporting 2 clickable icons


zarahzachz:
@seancdavis I added the `.connect-layout-container` to the searchbar component CSS. I tried adding this to the neighbors component CSS (which houses the whole app) but Angular doesn't allow for component-level CSS to be shared by other components - even if they're nested within each other.

This prevents the searchbar layout from looking goofy in Maestro.

Ready for review.

PR #445, opened 2017-07-05
defect/DE3831 Set clear text button to unhidden if there is text in the search bar

vchtelmakh:


vchtelmakh:
Good call on the unit tests, colleague.

PR #446, opened 2017-07-06
Defect/de3819 keep data around

JoeKer1:


src/app/components/create-group/create-group-data.service.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                // { provider: ServiceName, useClass: FakeServiceClass },
+                // { provider: ServiceName, useFactory: fakeServiceFactory, deps: [] },
+            ]
+        });
+    });
+
+    // you can also wrap inject() with async() for asynchronous tasks
+    // it('...', async(inject([...], (...) => {}));
+
+    fit('should create the create group service',
```
vchtelmakh: Unfit tests.

src/app/components/create-group/create-group-data.service.ts
```diff
@@ -1,15 +1,33 @@
...
+            return this.lookupService.getCategories()
+            .do((cats: Category[]) => {
+                this.categories = cats;
+                this.pageOneInitialized = true;
+            });
+        } else {
+            return Observable.of(this.categories);
+        }
+    }
+
```
vchtelmakh: Shouldn't categories come from a route resolve if they are always required?

vchtelmakh: Should we consider preloading categories in the component that requires them, in a route guard?

src/app/components/create-group/create-group-data.service.spec.ts
JoeKer1: dang it i knew i forgot something

src/app/components/create-group/create-group-data.service.ts
JoeKer1: I suppose we could move this logic into a guard but then we'd have to load it into the service on submit of the form and then the guard would have to get the data from the service. Shouldn't be a big deal if that's what we wanna do.

vchtelmakh: " to load it into the service on submit of the form and then the guard would have to get the data from the service or the backend depending on if it has been initialized" We could pass the cats to whatever service method uses them.

src/app/components/create-group/page-1/create-group-page-1.component.ts
```diff
@@ -1,31 +1,60 @@
...
+        }
+    }
+
+    public onSubmit(form) {
+        this.isSubmitted = true;
+        if (form.valid) {
+            // Do Something
+        } else {
+            // Do something else
+        }
```
vchtelmakh: Do we need to do something here, or are these placeholders for when we get more pages?

JoeKer1: Placeholders until we get more pages

config/webpack.test.js
```diff
@@ -51,6 +52,14 @@ module.exports = function(options) {
...
         systemvars: true
       }),
+
+      new webpack.ContextReplacementPlugin(
+         // The (\\|\/) piece accounts for path separators in *nix and Windows
+         /angular(\\|\/)core(\\|\/)@angular/,
+         helpers.root('./src'), // location of your src
+         { }
+       ),
+
```
Doug-Shannon: Nice!

src/app/components/create-group/page-1/create-group-page-1.component.ts
```diff
@@ -1,31 +1,60 @@
...
+
+    constructor(private state: StateService,
+                private createGroupService: CreateGroupService) { }
 
     ngOnInit() {
+        this.state.setLoading(true);
         this.state.setPageHeader('start a group', '/create-group');
-        this.initializeCategories();
+        this.groupCategoryForm = new FormGroup({});
+        this.createGroupService.initializePageOne()
```
Doug-Shannon: if you want me to be nitpicky, I don't like trailing dot notation.

JoeKer1: I do want you to be nitpicky. I don't like it either but it helps me keep track of like finally().subscribe() etc. 
Do you prefer: 
`this.createGroupService.initializePageOne().finally(
() => {
  this.state.setLoading(false);
}).subscribe(cats => {
  this.initializeCategories(cats);
});`

PR #449, opened 2017-07-10
re-implements cache in pinDetailsService, also fixes some logic errors

Doug-Shannon:
~~DO NOT MERGE~~ We can merge once everyone takes a look

### DIFF
--- a/src/app/services/state.service.ts
+++ b/src/app/services/state.service.ts
@@ -52,6 +52,10 @@ export class StateService {
     this.emitMyStuffChanged();
   }
 
+  public isMapViewSet(): boolean {
+    return this.savedMapView != null;
+  }
+
   public setMapView(mv: MapView) {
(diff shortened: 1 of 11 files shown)


src/app/services/state.service.ts
```diff
@@ -52,6 +52,10 @@ export class StateService {
     this.emitMyStuffChanged();
   }
 
+  public isMapViewSet(): boolean {
+    return this.savedMapView != null;
+  }
+
```
vchtelmakh: Is this initialized to  undefined: `public savedMapView: MapView;`

JoeKer1: this.savedMapView != null === !!this.savedMapView

!= null checks for both null and undefined. it's a weird javascript thing. If it were !== then it would not check for both.

JoeKer1:
removed requested changes

PR #451, opened 2017-07-11
Defect/de3795 group details formatting

zarahzachz:
Groups categories should have a space after ":", separating the group name from the group details.

Test by visiting the list view on `/connect` (categories should be visible here by default) and clicking into any list-entry to view gathering details categories.

No corresponding PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "angulartics2": "2.2.1",
     "copyfiles": "1.0.0",
     "core-js": "2.4.1",
-    "crds-styles": "0.9.0",
+    "crds-styles": "crdschurch/crds-styles#development",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
     "angulartics2": "2.2.1",
     "copyfiles": "1.0.0",
     "core-js": "2.4.1",
-    "crds-styles": "0.9.0",
+    "crds-styles": "crdschurch/crds-styles#development",
```
seancdavis: @zarahzachz: Normally I'd say hold off on doing this until we actually need updated style changes. But in this case we know we're making style changes for this app throughout this sprint, so I'm not worried about this change.

PR #456, opened 2017-07-12
Feature/us8872 group form first two pages

JoeKer1:


src/app/components/create-group/page-1/create-group-page-1.component.html
```diff
@@ -45,4 +47,7 @@ <h3 class="media-heading">Group Category</h3>
             </form>
         </div>
     </div>
-</div>
\ No newline at end of file
+</div>
+<!-- <div>
+    <pre>{{createGroupService.categories | json}}</pre>
```
sseissiger: Are we ready to delete this, or do you need it around for the final save story?

src/app/components/create-group/page-2/create-group-page-2.component.html
```diff
@@ -0,0 +1,99 @@
...
+                    <button type="submit" class="btn btn-secondary btn-block-mobile pull-right">Next</button>
+                    <button type="button" (click)="back()" class="btn btn-gray btn-default btn-block-mobile">Back</button>
+                </div>
+            </form>
+        </div>
+    </div>
+</div>
+<!-- <div>
+    <pre>{{createGroupService.meetingTimeType}}</pre>
+    <pre>{{meetingTimeForm.valid}}</pre>
```
sseissiger: remove commented code

src/app/components/create-group/page-2/create-group-page-2.component.ts
```diff
@@ -0,0 +1,91 @@
...
+
+@Component({
+    selector: 'create-group-page-2',
+    templateUrl: './create-group-page-2.component.html',
+})
+export class CreateGroupPage2Component implements OnInit {
+    public meetingTimeForm: FormGroup;
+    private isSubmitted: boolean = false;
+    private daysOfTheWeek: LookupTable[] = [];
+    private meetingFrequencies = [{
```
sseissiger: Should these come from MP? Instead of hard coded.  Little tricky since we are not using all the data from Meeting_Frequencies table.  Not sure what we are doing today in Group Tool.  
Can you at least pull out into constants because we will need this same data in filters.  Thanks

src/app/components/create-group/page-1/create-group-page-1.component.html
JoeKer1: was going to wait to remove these until the whole form is done.

src/app/components/create-group/page-2/create-group-page-2.component.ts
JoeKer1: I pulled that directly from what we do today in group tool.

src/app/components/create-group/page-1/create-group-page-1.component.html
JoeKer1: These also still exist commented out in current group tool :)

PR #457, opened 2017-07-13
-Fix button alignment

JoeKer1:
-Save categories to model properly for entire flow submission
-Add missing unit tests that I forgot to write.

### DIFF
--- a/src/app/components/create-group/create-group-data.service.ts
+++ b/src/app/components/create-group/create-group-data.service.ts
@@ -1,15 +1,17 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 
-import { Address, Group } from '../../models';
+import { Address, Group, Attribute } from '../../models';
 import { Category } from '../../models/category';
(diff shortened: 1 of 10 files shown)


src/app/components/create-group/create-group-data.service.ts
```diff
@@ -1,15 +1,17 @@
...
 import { Category } from '../../models/category';
 import { LookupService } from '../../services/lookup.service';
+import { groupCategoryAttributeTypeId } from '../../shared/constants';
 
 @Injectable()
 export class CreateGroupService {
     public meetingTimeType: string = 'specific';
     private pageOneInitialized: boolean = false;
     public categories: Category[] = [];
+    private selectedCategories: Category[] = [];
```
Doug-Shannon: We spoke about the categories & selectedCategories.  Its a little weird to me to merge the model and the lookup data into one spot, then have just the model as well.  If this service starts getting complicated, this could be a spot we simplify.

PR #458, opened 2017-07-13
US8875 updated participant names

simeonsena:


src/app/components/list-entry/list-entry.component.ts
```diff
@@ -70,6 +70,17 @@ export class ListEntryComponent implements OnInit {
...
+  public isMySmallGroup() {
+    return this.type === pinType.SMALL_GROUP && this.contactId === this.currentContactId;
+  }
+
+  public formatLeaderName() {
+   // if (this.isMyGathering) {
+    //  return (this.firstName + ' ' + this.lastName);
+   // } else {
+      return this.firstName + ' ' + this.lastName.charAt(0) + '.';
+   // }
```
vchtelmakh: Do we not need the check? If so, delete the comments.

src/app/components/pin-details/gathering/gathering.component.ts
```diff
@@ -84,7 +85,8 @@ export class GatheringComponent implements OnInit {
 
           this.participantService.getCurrentUserGroupRole(this.pin.gathering.groupId).subscribe(
             role => {
-              if (role !== GroupRole.NONE) {
+              let isInGroup: boolean = role !== GroupRole.NONE;
```
JoeKer1: do we use isInGroup anywhere else? if not I would change it back, but that may be personal preference

src/app/components/pin-details/gathering/gathering.component.ts
```diff
@@ -125,6 +128,15 @@ export class GatheringComponent implements OnInit {
     this.router.navigate([contactLeaderOfThisGroupPageUrl]);
 
   }
+ 
+  private getAdjustedLeaderNames(leaders: Participant[], isUserParticipant: boolean): string[] {
+     let adjustedLeaderNames: string[] = [];
+     leaders.forEach((leader) =>{
+       let adjustedName: string =  isUserParticipant ? `${leader.nickName} ${leader.lastName}` : `${leader.nickName} ${leader.lastName.slice(0,1) + "."}`;
```
JoeKer1: instead of doing the + '.' here you could just do 

`'${leader.nickName} ${leader.lastName.slice(0,1)}.'`

PR #471, opened 2017-07-18
Feature/us8971 show min 1 group

harimau777:
When the user searches the map's zoom is now adjusted to include the target
number of pins (currently 1 for groups and 10 for connect) when possible.
However, it is set to have a maximum and minimum zoom of 3 and 15 respectively.
In order to solve the problem with pins appearing right on the edge of the map,
after the proper zoom is calculated, .25 is subtracted in order to zoom out
slightly farther (so the actual range of allowable zooms is 2.75 to 14.75).
We can adjust this extra zoom out later if we desire.

The zoom calculation is handled in google-map.service.ts, and I added tests for
the zoom calculation in google-map.service.spec.ts.  These tests only cover the
other functionality of the GoogleMapService (such as calculating bounding boxes)
indirectly.  It may be a good idea to add tests for those features in the
future.  In order to facilitate testing, I added a public method,
getPinsInBounds, which may be useful in other situations.

The current algorithm for adjusting the zoom starts with a zoom of 9 and then
uses recursion to repeatedly increase or decrease the zoom by 1 until it finds
the maximum zoom that contains the target number of pins.  The starting zoom was
chosen to be halfway between 3 and 15; we may want to look at whether we should
use a different starting zoom.  A more efficient algorithm would be to instead
adjust by half of difference between the current zoom and the min or max zoom
(depending on whether the zoom needs to be adjusted up or down).  In effect, we
would be doing a binary search for the proper zoom.

Finally, I significantly refactored the code in google-map.service.ts.


src/app/services/google-map.service.ts
```diff
@@ -54,80 +52,56 @@ export class GoogleMapService {
...
-        // width: document.documentElement.clientWidth,
-        // height: document.documentElement.clientHeight,
-        // lat: lat,
-        // lng: lng
-        // };
-        // return this.calculateBestZoom(bounds, zoom, pins, viewtype);
+    const mapParams: MapView = new MapView('zoomCalcView', lat, lng, initialZoom);
+    const popTarget = this.getPopTarget(pins, viewtype);
+
+    return this.calculateBestZoom(mapParams, pins, popTarget, {}) - .25;  // Subtract .25 so that results are not on edge of screen
```
vchtelmakh: Should be constant -  screenEdgeBuffer or something. I can't tell what this does, as far as I know zoom is an integer, so I don't know what happens when a decimal is subtracted from it.

src/app/services/google-map.service.ts
```diff
@@ -54,80 +52,56 @@ export class GoogleMapService {
...
 
-  // zero in on the zoom that's closest to the target pin count without going under
-  private calculateBestZoom(bounds: Object, zoom: number, pins: Pin[], viewtype: string, pops: Object = {}): number {
-    let popTarget;
+  // Returns the target number of search results based on whether this is a connect or group app.
+  private getPopTarget(pins: Pin[], viewtype: string): number {
     if (viewtype === 'world') {
-      popTarget = 10;
+      if (this.appSettings.isConnectApp()) {
+        return 10;
```
vchtelmakh: 10 should be a constant: 'minNumberOfConnectPinsOnMap" or something like that.

src/app/services/google-map.service.ts
```diff
@@ -54,80 +52,56 @@ export class GoogleMapService {
...
+      } else {
+        return 1;
```
vchtelmakh: Ditto with groups.

src/app/services/google-map.service.ts
```diff
@@ -54,80 +52,56 @@ export class GoogleMapService {
...
+  }
 
-    let pop = this.countPopAtZoom(bounds, zoom, pins, pops);
+  // zero in on the zoom that's closest to the target pin count without going under
+  private calculateBestZoom(mapParams: MapView, pins: Pin[], popTarget: number, pops: Object = {}): number {
+    const pop: number = this.countPopAtZoom(mapParams, pins, pops);
     if (pop < popTarget) {
-      if (zoom <= 3) {
+      if (mapParams.zoom <= 3) {
         return 3;
```
vchtelmakh: Is three just as far as we ever want to zoom out? Also a constant for readability.

src/app/services/google-map.service.ts
```diff
@@ -144,64 +118,63 @@ export class GoogleMapService {
 
     // vadjust compensates for the local distortion of the Mercator projection.
 
-    let zoom = mapParams.zoom - 1; //Why adjust with "-1"? No one knows - magic danderson code.
-    let docWidth = document.documentElement.clientWidth;
-    let docHeight =  document.documentElement.clientHeight;
-    let centerLat = mapParams.lat;
-    let centerLng = mapParams.lng;
+    const zoom = mapParams.zoom - 1; // Why adjust with "-1"? No one knows - magic danderson code.
```
vchtelmakh: Maybe this is here with the same intent as you adjusting the zoom level. Not something to really look into deeply.

plachmann:
Handled with a different PR

PR #472, opened 2017-07-18
us9075 respond to search error

plachmann:


src/app/components/neighbors/neighbors.component.ts
```diff
@@ -138,16 +142,29 @@ export class NeighborsComponent implements OnInit, OnDestroy {
...
+        this.state.setLastSearchSearchString(lastSearchString);
         this.state.setLoading(false);
-        this.goToNoResultsPage();
+        this.goToErrorPage();
       });
-
   }
 
+  private goToErrorPage() {
+    let errorText = 'Oops, looks like there was a problem. Please try again.';
```
JoeKer1: I think you can do this.blandPageService.goToDefaultError(); or something like that. I think it does the same thing.

PR #488, opened 2017-07-21
Refactoring

harimau777:
Replaced != with !== in multiple locations


src/app/components/create-group/create-group-data.service.ts
```diff
@@ -98,7 +98,7 @@ export class CreateGroupService {
                             category.categoryId, null, 0, attributeTypes.GroupCategoryAttributeTypeId,
                             null, null);
 
-        if (category.attribute != null) {
+        if (category.attribute !== null) {
```
JoeKer1: != here was on purpose to handle undefined and null.

JoeKer1: You can repeat this comment throughout this pull request. :)

src/app/services/pin.service.ts
```diff
@@ -92,11 +92,11 @@ export class PinService extends SmartCacheableService<PinSearchResultsDto, Searc
...
         if (aPin.pinType === pinIdentifier.type) {
           if (pinIdentifier.type === pinType.PERSON) {
-            return (aPin.participantId == pinIdentifier.id);  // need == not === here b/c have string and number		
+            return (aPin.participantId === pinIdentifier.id);  // need == not === here b/c have string and number
           } else if (pinIdentifier.type === pinType.GATHERING) {
-            return (aPin.gathering.groupId == pinIdentifier.id); // need == not === here b/c have string and number		
+            return (aPin.gathering.groupId === pinIdentifier.id); // need == not === here b/c have string and number
           } else if (pinIdentifier.type === pinType.SMALL_GROUP) {
-            return (aPin.gathering.groupId == pinIdentifier.id); // need == not === here b/c have string and number		
+            return (aPin.gathering.groupId === pinIdentifier.id); // need == not === here b/c have string and number
```
JoeKer1: not 100% this will work. Once upon a time, sometimes it would be a number and sometimes it would be a string. hence the comments you should get rid of if you keep these changes.

Doug-Shannon:
I think this is a lot of risk for very little reward.  Some places we consciously chose == over === or != over !== and changing them after the fact could result in unexpected and uncaught errors.

src/app/components/list-entry/list-entry.component.html
```diff
@@ -63,8 +63,8 @@ <h3 class="media-heading text-uppercase font-family-condensed-extra">{{formatNam
       </span>
     </h3>
     <p class="media-meta">
-      <span *ngIf="pin?.gathering?.meetingDay != null">{{pin?.gathering?.meetingDay}} at {{ getMeetingTime() | date:'shortTime' }}, {{pin?.gathering?.meetingFrequency}}</span>
-      <span *ngIf="pin?.gathering?.meetingDay == null">Flexible Meeting Time</span>
+      <span *ngIf="pin?.gathering?.meetingDay !== null">{{pin?.gathering?.meetingDay}} at {{ getMeetingTime() | date:'shortTime' }}, {{pin?.gathering?.meetingFrequency}}</span>
```
Doug-Shannon: This span would be displayed if meetingDay was undefined but not if it was null.  so you would have to put !== null && !== undefined.

PR #498, opened 2017-07-25
Feature/us8921 start a group page6

JoeKer1:
This can actually be merged without the backend changes. Just don't expect your pin to show up in aws until it syncs. Which is faster now right?


src/app/components/address-form/address-form.component.ts
```diff
@@ -17,6 +17,7 @@ export class AddressFormComponent implements OnInit {
   @Input() groupName: string;
   @Input() address: Address;
   @Input() isFormSubmitted;
+  @Input() errorClasses: string = 'error help-block';
```
vchtelmakh: Why is this an input?

JoeKer1: because I tried to reuse this form on the profile section but I couldn't because group stores zip in zip and profile stores it in postalCode because we don't need no consistency. So I needed the error messages to look differently. I can take out the changes to this if you want.

src/app/components/create-group/create-group-data.service.ts
```diff
@@ -81,30 +99,85 @@ export class CreateGroupService {
...
+    public reset(): void {
+        this.pageOneInitialized = false;
+        this.pageSixInitialized = false;
+        this.profileData = {};
+        this.group = Group.overload_Constructor_CreateGroup(0);
+        this.selectedAgeRanges = [];
+        this.selectedCategories = [];
+        this.selectedGroupGenderMix = Attribute.constructor_create_group();
+        this.categories = [];
+        this.meetingTimeType = 'specific';
```
vchtelmakh:
```export class GroupMeetingScheduleType {
  SPECIFIC_TIME_AND_DATE: string;
  FLEXIBLE: string;

  constructor() {
    this.SPECIFIC_TIME_AND_DATE = 'specific';
    this.FLEXIBLE = 'flexible';
  }
}

export const groupMeetingScheduleType: any = new GroupMeetingScheduleType();```

JoeKer1: Yes you should've done this when you added that. ;)

PR #512, opened 2017-07-26
Feature/us9056 cleaning up labels

harimau777:


src/app/components/pin-details/gathering/gathering.html
```diff
@@ -25,7 +25,7 @@
         {{descriptionToDisplay}}
         <a *ngIf="!doDisplayFullDesc" (click)="expandGroupDescription()" class="pointer">More</a>
 
-        <span *ngIf="doDisplayFullDesc">
+        <!-- <span *ngIf="doDisplayFullDesc">
```
plachmann: if this code is un-needed you can go ahead and delete it. It increases readability.

PR #522, opened 2017-07-31
Feature/us9094 filter search on day time frequency v3

vchtelmakh:


src/app/shared/constants.ts
```diff
@@ -121,16 +181,31 @@ export const maxZoom: number = 15;          // The maximum zoom before zoomAdjus
...
-            meetingFrequencyId: 1,
-            meetingFrequencyDesc: 'Every week'
-        }, {
-            meetingFrequencyId: 2,
-            meetingFrequencyDesc: 'Every other week'
-        }, {
-            meetingFrequencyId: 8,
-            meetingFrequencyDesc: 'Every month'
-        }];
+  meetingFrequencyId: 1,
```
JoeKer1: I get it, but this just feels weird.

src/app/models/simple-selectable.ts
```diff
@@ -0,0 +1,8 @@
+export class SimpleSelectable {
```
JoeKer1: Can/Did you replace existing models with this one?

vchtelmakh: I didn't, though I probably could. I think it should be done in a refactor story though. I am making a refactor story to remove redundancy in the filter component, it's silly.

src/app/shared/constants.ts
vchtelmakh: If you mean the names being passed in - I am using them elsewhere, so I want to eliminate hard-coding in more than one place.

PR #525, opened 2017-07-31
Feature/us9270 participant edit

zarahzachz:
As a group leader, 
I can view details about a participant in my group, remove them from the group, or assign them a role within the group. We need to update markup on this interface. 

https://invis.io/68BRD8RXB#/237976894_05M-01b-My-Group-Participant

No corresponding PRs


zarahzachz:
@seancdavis I took out the extra spacing between the save button and last paragraph. All spacing should look about the same.

Ready for review.

PR #528, opened 2017-08-01
US9304 - Groups Header Redux

seancdavis:
Given a crds.net user 
When directed to the Groups
Then I should see the appropriate header for Groups within the Finder application

When directed to Connect
Then I should see the appropriate header for the Connect application

---

The biggest thing to check here is that the buttons work as expected when you are not on the map or list view.

---

Corresponds with crdschurch/crds-maestro#205.


src/app/app.component.ts
```diff
@@ -77,6 +77,9 @@ export class AppComponent implements OnInit {
       this.defaultToGroupAppType();
     }
 
+    // Uncomment the following line to force Connect app when running outside
+    // Maestro. But be sure not to commit!
+    // this.appsettings.setAppSettings(appType.Connect);
```
tcmacdonald: I'm not following your comment. What does this line do @seancdavis ?

PR #554, opened 2017-08-03
DE4042: Fix Proximity

vchtelmakh:
----


JoeKer1:
Was the number pipe not working?

vchtelmakh:
Works fine. Pretty sure it doesn't work on strings though, which is what my method returns.

JoeKer1:
You could still use the pipe in the controller

PR #564, opened 2017-08-07
Privacy not set message no longer appears when other fields are blank

harimau777:


src/app/components/create-group/page-5/create-group-page-5.component.ts
```diff
@@ -43,7 +43,6 @@ export class CreateGroupPage5Component implements OnInit {
     if (form.valid && this.createGroupService.group.availableOnline != null) {
       this.router.navigate(['/create-group/page-6']);
     } else {
-      this.groupVisibilityInvalid = true;
```
vchtelmakh: What did it do in the first place? Are you sure we don't need it for form validation?

PR #570, opened 2017-08-09
Feature/us8761 non active leader view

sseissiger:


src/app/components/stuff-not-found/stuff-not-found.html
```diff
@@ -1,5 +1,13 @@
...
+  
+  <!-- change to standard styles if isApprovedLeader -->
+  <button *ngIf="!isApprovedLeader" class="btn btn-default btn-outline btn-block-mobile" (click)="onFindAGroupClicked()">Find a group</button>
+  <button *ngIf="isApprovedLeader" class="btn btn-primary btn-block-mobile" (click)="onFindAGroupClicked()">Find a group</button>
+  
+  <button *ngIf="isApprovedLeader" class="btn btn-primary btn-block-mobile" (click)="onGroupResourcesClicked()">Group resources</button>
+  
+  <!-- change to default styles if isApprovedLeader -->
+  <button *ngIf="isApprovedLeader" class="btn btn-default btn-outline btn-block-mobile" routerLink="/create-group">Start a group</button>
+  <button *ngIf="!isApprovedLeader" class="btn btn-primary btn-block-mobile" routerLink="/create-group">Start a group</button>
```
vchtelmakh: Please use conditional styles here - as far as I can tell that's the only difference, and an `*ngIf* implies logical differences. 

https://stackoverflow.com/questions/35269179/angular-2-conditional-class-with-ngclass

PR #574, opened 2017-08-10
Defect/de4032 switch to list view when filtering out physical groups

harimau777:
https://rally1.rallydev.com/#/66097202244d/detail/defect/140949607744

### DIFF
--- a/src/app/components/filters/filters.component.ts
+++ b/src/app/components/filters/filters.component.ts
@@ -18,7 +18,7 @@ import { MeetingFrequencyComponent } from './meeting-frequency/meeting-frequency
 
 import { PinSearchRequestParams } from '../../models/pin-search-request-params';
 
-import { listViewType } from '../../shared/constants';
+import { ViewType } from '../../shared/constants';
 
 
 @Component({
@@ -64,7 +64,7 @@ export class FiltersComponent implements OnInit {
 
     // Switch to list view if the user is searching for only online groups:
     if (this.onlineOrPhysicalGroupComponent.getIsVirtualGroup()) {
-      this.state.setCurrentView(listViewType);
+      this.state.setCurrentView(ViewType.LIST);
     }
 
     let filterString: string = this.filterService.buildFilters();
--- a/src/app/components/filters/filters.component.ts
+++ b/src/app/components/filters/filters.component.ts
@@ -46,7 +46,7 @@ export class FiltersComponent implements OnInit {
                private pinService: PinService,
                private state: StateService ) {}
 
-  ngOnInit() {
+  ngOnInit(): void {
     let savedSearch = this.state.lastSearch;
     this.locationFormGroup = new FormGroup({
         location: new FormControl(this.location, []),
--- a/src/app/components/filters/filters.component.ts
+++ b/src/app/components/filters/filters.component.ts
@@ -15,7 +15,6 @@ import { OnlineOrPhysicalGroupComponent } from './online-or-physical-group/onlin
 import { MeetingTimeComponent } from './meeting-time/meeting-time.component';
 import { MeetingDayComponent } from './meeting-day/meeting-day.component';
 import { MeetingFrequencyComponent } from './meeting-frequency/meeting-frequency.component';
-import { SearchBarComponent } from '../search-bar/search-bar.component';
 
 import { PinSearchRequestParams } from '../../models/pin-search-request-params';
 
@@ -45,8 +44,7 @@ export class FiltersComponent implements OnInit {
   constructor( private filterService: FilterService,
                private router: Router,
                private pinService: PinService,
-               private state: StateService,
-               private searchBar: SearchBarComponent ) {}
+               private state: StateService ) {}
 
   ngOnInit() {
     let savedSearch = this.state.lastSearch;
@@ -54,7 +52,7 @@ export class FiltersComponent implements OnInit {
         location: new FormControl(this.location, []),
     });
 
-    if ((savedSearch) && savedSearch.location != null){
+    if ((savedSearch) && savedSearch.location != null) {
       this.locationFormGroup.controls['location'].setValue(savedSearch.location);
     }
   }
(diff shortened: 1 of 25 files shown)


src/app/components/filters/filters.component.ts
```diff
@@ -37,14 +39,14 @@ export class FiltersComponent implements OnInit {
...
-
   public locationFormGroup: FormGroup;
   public location: string;
 
   constructor( private filterService: FilterService,
                private router: Router,
                private pinService: PinService,
-               private state: StateService) { }
+               private state: StateService,
+               private searchBar: SearchBarComponent ) {}
```
plachmann: we should not be injecting components like this. If you just need the setButtonText method in multiple components pull it out into an appropriate service and inject the service.

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -43,7 +42,7 @@ export class NeighborsComponent implements OnInit, OnDestroy {
     private state: StateService,
     private userLocationService: UserLocationService,
     private searchService: SearchService,
-    private filterService: FilterService,
+    private F: FilterService,
```
sseissiger: Please rename back to filter service not the letter F - thanks

src/app/components/search-bar/search-bar.component.spec.ts
```diff
@@ -32,6 +34,7 @@ describe('SearchBarComponent', () => {
     mockAppSettingsService = jasmine.createSpyObj<AppSettingsService>('appSettingsService', ['isConnectApp', 'isSmallGroupApp']);
     mockPinService = jasmine.createSpyObj<PinService>('pinService', ['emitPinSearchRequest']);
     mockFilterService = jasmine.createSpyObj<FilterService>('filterService', ['buildFilters']);
+    // mockStateService = jasmine.createSpyObj<StateService>('stateService', ['getCurrentView', 'myStuffStateChangedEmitter', 'setMyViewOrWorldView']);
```
plachmann: remove commented code

src/app/components/neighbors/neighbors.component.ts
```diff
@@ -53,24 +52,22 @@ export class NeighborsComponent implements OnInit, OnDestroy {
...
 
   public ngOnInit(): void {
-
     this.subscribeToListenForSearchRequests();
-    this.setViewToMapOrList(this.state.getCurrentView());
     this.runInitialPinSearch();
-
   }
 
-  private setViewToMapOrList(mapOrListView): void {
```
sseissiger: not sure why you need to make these changes here?
And if really needed - you need an access modifier and return type on the functions

src/app/components/neighbors/neighbors.component.html
```diff
@@ -4,7 +4,7 @@
 </app-search-bar>
 
 <div *ngIf="!this.state.getIsFilteredDialogOpen()">
-  <app-map [searchResults]="pinSearchResults" *ngIf="mapViewActive && !isMapHidden"></app-map>
```
sseissiger: not sure why you need to make these changes in this file - why renaming?

src/app/components/search-bar/search-bar.component.ts
```diff
@@ -81,8 +82,8 @@ export class SearchBarComponent implements OnChanges, OnInit {
     this.pinService.emitPinSearchRequest(pinSearchRequest);
   }
 
-  private setButtonText() {
-    this.buttontext = this.isMapHidden ? 'Map' : 'List';
+  public setButtonText() {
```
sseissiger: add return of void - thanks

src/app/components/neighbors/neighbors.component.html
plachmann: Getting the view (map or list) from a function to pull from state service.

src/app/components/neighbors/neighbors.component.spec.ts
```diff
@@ -143,42 +143,36 @@ describe('Component: Neighbors', () => {
...
+    expect(mockStateService.setCurrentView).toHaveBeenCalledWith(mapViewType);
+    expect(mockStateService.getMapView).toHaveBeenCalledTimes(1);               // Check that getMapView was not called again
+    expect(mockPinService.reSortBasedOnCenterCoords).toHaveBeenCalledTimes(1);  // Check that reSortBasedOnCenterCoords was not called again
   });
 
   it('should processAndDisplaySearchResults', () => {
     let searchResults: PinSearchResultsDto = MockTestData.getAPinSearchResults(1);
     this.component['pinSearchResults'] = searchResults;
-    this.component['mapViewActive'] = true;
+    // this.component['mapViewActive'] = true;
```
vchtelmakh: Remove commented code

src/app/components/neighbors/neighbors.component.ts
vchtelmakh: x2 - If I see "f" in the code, I have to go check what it is - this reduces clarity.

src/app/components/neighbors/neighbors.component.ts
vchtelmakh: x2 - set return type to boolean explicitly, despite it being implied by the function name

src/app/shared/constants.ts
```diff
@@ -146,6 +146,10 @@ export class AppType {
   }
 }
 
+// View Types:
+export const mapViewType: string = 'map';
+export const listViewType: string = 'list';
```
vchtelmakh: Shouldn't this be a mapViewType enum or object with these two options as properties?

src/app/components/neighbors/neighbors.component.ts
harimau777: Hmm, I must have accidentally highlighted and overwritten while I was changing something else.  I'm surprised it didn't cause any errors.

src/app/components/neighbors/neighbors.component.ts
harimau777: The problem was that the neighbors component was storing whether the view is map or list in one of its properties rather than checking the state service.  This meant that when other components or services changed the state, the changes weren't reflected in the neighbors component.

The isMapViewSet function was added because the neighbors component was storing the view as a boolean (true = map, false = list) while the state stores the view as a string (either Map or View).  Long term I think it would be better to consistently use a string (in case we want to add additional views in the future) but I added isMapViewSet in order to minimize the changes that need to be made to get this story done.

simeonsena: Putting my QA hat back on, what is the impact  of this change to other areas of the site?

PR #583, opened 2017-08-11
DE4072 - Btn dropdown not accessible

zarahzachz:
Add styles specific to the "contact participants" dropdown.

Crdschurch/crds-styles#197
Crdschurch/crds-styleguide#207


src/app/components/email-participants/email-participants.component.scss
```diff
@@ -0,0 +1,8 @@
+.dropdown .btn {
+  display: block;
+}
+
+.dropdown-menu {
+  position: relative;
+  padding-bottom: 1rem;
```
seancdavis: Why the extra padding at the bottom? It makes the menu look unbalanced.

screen shot 2017-08-14 at 9 49 10 am

zarahzachz: Oops... switched it to margin instead of padding.

PR #585, opened 2017-08-11
Feature/us9503 fauxdal cleanup

zarahzachz:
Clean up the fauxdals so they match the rest of the application.

First fauxdal is in the "Add a Participant" section in your group. 
Second is when a user hits the cancel link in the Create a Group section.

In both instances, text should match other fauxdals (light weight, centered, white, etc.) and buttons should be centered and go full-width on mobile-sized screens. Added some JS to the components to hide the scrollbar on `body` if the modal is open, and show it when modal is closed.

No corresponding PRs.


zarahzachz:
@seancdavis I resolved up my conflict. Ready for review.

PR #586, opened 2017-08-11
DE4082 - oops

zarahzachz:
Added markup to a bland style page so it matches other bland pages.

Good luck recreating! I got there by signing in and clicking the My Groups button in the header. This works some of the time... 

Here's how the page looks on my screen:
screen shot 2017-08-11 at 2 56 06 pm

No corresponding PRs


seancdavis:
I would usually suggest you use a content block here, but it wasn't part of the requirements and it wasn't there in the first place, so going to merge as is.

PR #600, opened 2017-08-16
DE4115

vchtelmakh:
Code was setting flag in each iteration (i.e. after first cat being set, instead of looping through every cat). Moved flag to trigger after loop.


simeonsena:
looks good

PR #625, opened 2017-08-23
us9091 social media icons

plachmann:


src/app/components/pin-details/gathering/gathering.component.ts
```diff
@@ -151,6 +151,14 @@ export class GatheringComponent implements OnInit {
...
   }
 
+  public showSocial(): boolean {
+    return !(this.state.getActiveGroupPath() === groupPaths.EDIT || this.state.getActiveGroupPath() === groupPaths.ADD);
+  }
+
+  public isPublicGroup(): boolean {
+    return this.pin.gathering.address !== undefined;
+  }
+
```
vchtelmakh: There might be a different way to do this:
AWS: 'groupavailableonline' field / `group.availableOnline` field on frontend. 

I am NOT sure these are the same thing.

plachmann: You are correct. I made the change

PR #626, opened 2017-08-23
DE4147 - Try Group Confirmation TLC

seancdavis:
Go to /groupsv2/search.
Log in using tester11@gmail.com / password.
Go to list view.
Click on any group that tester11 isn't already a member of.
Click Try the group.


zarahzachz:
@seancdavis I think this fauxdal needs a style override. It had the double-scrollbar effect and the background is scrollable. 
screen shot 2017-08-24 at 4 47 03 pm

I think the way I've gone about fixing this is by adding style overrides to the component effecting the `overflow-y` on the fauxdal itself and I *think* the body of the page. But if you think this is 👍 then that's fine by me and I'll merge.

PR #636, opened 2017-08-25
Single quotes in search string are now URL encoded to %27

harimau777:
https://rally1.rallydev.com/#/66097202244d/detail/defect/146571902544

### DIFF
--- a/src/app/components/search-bar/search-bar.component.ts
+++ b/src/app/components/search-bar/search-bar.component.ts
@@ -65,11 +65,9 @@ export class SearchBarComponent implements OnChanges, OnInit {
     this.state.myStuffActive = false;
     this.state.setMyViewOrWorldView('world');
     this.state.setIsFilterDialogOpen(false);
-
-    // Remove apostrophes from search string:
-    search = search.split('')
-    .filter((letter) => letter !== '\'')
-    .join('');
+    this.state.searchBarText = search;
+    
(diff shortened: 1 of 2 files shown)


src/app/components/search-bar/search-bar.component.ts
```diff
@@ -65,11 +65,9 @@ export class SearchBarComponent implements OnChanges, OnInit {
...
     this.state.setMyViewOrWorldView('world');
     this.state.setIsFilterDialogOpen(false);
-
-    // Remove apostrophes from search string:
-    search = search.split('')
-    .filter((letter) => letter !== '\'')
-    .join('');
+    this.state.searchBarText = search;
+    
+    search = search.replace(/'/g, '%27');  // Escape single quotes in the search string
```
JoeKer1: Instead of replacing a specific instance of something that breaks search, should we URL encode it? does double quotes or ; or # or / or \ do something bad to the searches?

harimau777: The URL encoder built into JavaScript does not encode single parentheses.  We verified that several other non-alphanumeric characters do not cause an error; however, we didn't confirm that they would actually match a group that had that character in the title (mainly because I'm not sure there are any; at least until I start my group called 1337 H@XZ0R5!)

JoeKer1: lol :)

PR #637, opened 2017-08-25
DE4157

vchtelmakh:
Path was relative to parent path. When we came from email link, path was incorrect, so navigation broke.

### DIFF
--- a/src/app/components/pin-details/gathering/gathering.html
+++ b/src/app/components/pin-details/gathering/gathering.html
@@ -82,7 +82,7 @@
     <hr class="push-quarter-top flush-bottom">
     <div *ngFor="let participant of pin.gathering.Participants">
       <div class="border-bottom">
-        <participant-card [participant]="participant" [pinParticipantId]="pin.participantId"></participant-card>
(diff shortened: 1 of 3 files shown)


JoeKer1:
Could we discuss? I'm not sure what the issue was. It used to work for both groups and connect before. Also not sure what groupCardIsDisplayedOn but I'm probably dense.

src/app/components/pin-details/gathering/gathering.html
```diff
@@ -82,7 +82,7 @@ <h3 class="component-header font-size-h5 text-uppercase flush-bottom">Participan
     <hr class="push-quarter-top flush-bottom">
     <div *ngFor="let participant of pin.gathering.Participants">
       <div class="border-bottom">
-        <participant-card [participant]="participant" [pinParticipantId]="pin.participantId"></participant-card>
+        <participant-card [participant]="participant" [pinParticipantId]="pin.participantId" [groupCardIsDisplayedOn]="pin.gathering"></participant-card>
```
JoeKer1: What does group card is displayed on mean? Do we need the group id or something?

PR #643, opened 2017-08-28
DE4166 - Clickable Icon Area

seancdavis:
Fix clickable area on pin details icons


zarahzachz:
This looks good! Gold star!!! 🏅

PR #670, opened 2017-09-07
Defect/de4244 my groups view not scrollable

vchtelmakh:


src/app/app.module.ts
```diff
@@ -114,6 +114,7 @@ import { IPService } from './services/ip.service';
 import { LocationService } from './services/location.service';
 import { LoginRedirectService } from './services/login-redirect.service';
 import { LookupService } from './services/lookup.service';
+import { MiscellaneousService } from './services/miscellaneous-service';
```
JoeKer1: I have not scrolled down to see what is in this service but this name

src/app/services/miscellaneous-service.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class MiscellaneousService {
+
+  constructor() {
+  }
+
+  public reEnableScrollingInCaseFauxdalDisabledIt(): void {
+    document.querySelector('body').style.overflowY = 'scroll';
```
JoeKer1: When do we need to do this? Do we know when we should use this method?

src/app/app.module.ts
vchtelmakh: It beats "Random **** service" imo

src/app/services/miscellaneous-service.ts
vchtelmakh: This method is used when a fauxdal is opened and then we try to go back to a normal page. The fauxdals set overflowY to no-scroll value. This might need to be used in a couple other places throughout the site, hence the service.

src/app/app.module.ts
JoeKer1: rofl

src/app/services/miscellaneous-service.ts
JoeKer1: Yeah I get why it's in a service, had to do it in more than one place. Is there another fix for this? Do fauxdals need to set overflowY to no-scroll? Or is there something we can do on onDestroy of a faudal?

vchtelmakh: That could be a solution - I would need to look at it more. `Do fauxdals need to set overflowY to no-scroll?` - talked to Ample about it, they said it was set in CSS somehow.

Doug-Shannon: Two things - Touching the UI from a service is a really bad idea and breaks MVC paradigm.  I would strongly suggest we do this in another way.  Maybe we can fix it by using better and more specific css selectors in the fauxdal?  Also, I loathe the idea of a miscellaneous service - its not a huge deal right now, but in a year, this thing will be bloated and doing a billion different things.

vchtelmakh: I am going to either touch it from the controller or the service, unless someone has any better ideas. As far as the misc service; it's going to be called dom-manipulation service or something along those lines otherwise.

vchtelmakh: @JoeKer1 @Doug-Shannon Multiple components use the fauxdal, so onDestroy would have to be implemented on every component where the template html is wrapped in fauxdals. This won't be any cleaner then what I have now. Let me know any further concerns.

Doug-Shannon: @vchtelmakh I think it would only be required in the onDestroy of the bland page controller itself.  Honestly, having the document selector in that controller is also not optimal.

vchtelmakh: @Doug-Shannon It's not always the bland page component - for example, it happens in `end-group-confirmation.component.ts` and anywhere else something is wrapped in a fauxdal.

JoeKer1:
I do not believe we should have merged this.

PR #678, opened 2017-09-12
US9747 get single content category

plachmann:


JoeKer1:
This will break a lot of things. We use content blocks from all the categories listed in the for root of the content block module

JoeKer1:
I think we need to keep general even after the script

JoeKer1:
Whichever cat the generalError is in

plachmann:
Joe. See the SS-CMS PR that goes with this.

PR #687, opened 2017-09-14
Feature/update google maps

JoeKer1:
Updated to latest version of angular google maps. Everything seems to work with the agm/js-clusterer package as well. Should we still use our directive or this one?


src/app/components/map/map.component.ts
```diff
@@ -24,6 +24,14 @@ import { MapView } from '../../models/map-view';
 export class MapComponent implements OnInit {
 
   @Input() searchResults: PinSearchResultsDto;
+  private styles = [{
+          url: '//crds-cms-uploads.s3.amazonaws.com/connect/CLUSTER.svg',
+          height: 50,
+          width: 50,
+          textColor: '#fff'
+        }];
```
vchtelmakh: Probably a const, but ¯\_(ツ)_/¯

src/app/components/map/map.component.ts
```diff
@@ -24,6 +24,14 @@ import { MapView } from '../../models/map-view';
...
+  private maxZoom = 17;
```
harimau777: The maximum we use when calculating the zoom is 15, do we want to use the same value here for the maximum the user can manually zoom out?

Should any of these settings be added to the constants file?

src/app/components/map/map.component.ts
```diff
@@ -24,6 +24,14 @@ import { MapView } from '../../models/map-view';
 export class MapComponent implements OnInit {
 
   @Input() searchResults: PinSearchResultsDto;
+  private styles = [{
```
harimau777: Should we/can we add this to one of the css files?

src/app/components/map/map.component.ts
```diff
@@ -24,6 +24,14 @@ import { MapView } from '../../models/map-view';
...
+          url: '//crds-cms-uploads.s3.amazonaws.com/connect/CLUSTER.svg',
+          height: 50,
+          width: 50,
+          textColor: '#fff'
+        }];
+  private maxZoom = 17;
+  private averageCenter = true;
```
harimau777: The AOT requires properties that are used by a template to be public.  Do we want to start doing this even though the AOT isn't ready yet?

src/app/components/map/map.component.ts
JoeKer1: Yeah I just wanted to get it working first :D

src/app/components/map/map.component.ts
JoeKer1: I just used what was already there in our directive, I'm not sure what it should be. Should probably be in constants though.

src/app/components/map/map.component.ts
JoeKer1: I read an issue last night that stated that none of the angular google maps packages support AoT yet, but that's probably not a bad idea if we have to do it for AoT in the future.

JoeKer1: https://github.com/SebastianM/angular-google-maps/issues/1143

JoeKer1:
Right now  this breaks labels. So we may need to go back to our old clusterer for now.

PR #694, opened 2017-09-15
DE4278: Wow such complex very difficulty

vchtelmakh:


JoeKer1:
no tests need updated / created for this?

vchtelmakh:
@JoeKer1 We have some prod issues atm, so no, but I can write some after I take care of that.

PR #707, opened 2017-09-21
Feature/us9938 move to boiler plate

JoeKer1:
This PR will move the finder app foundation to the angular boilerplate project
[Must go with this pr](https://github.com/crdschurch/crds-maestro/pull/238)


JoeKer1:
[Must go with this pr](https://github.com/crdschurch/crds-maestro/pull/238)

package.json
```diff
@@ -1,106 +1,73 @@
...
-    "copyfiles": "1.0.0",
-    "core-js": "2.4.1",
-    "crds-styles": "0.13.0",
-    "iframe-resizer": "3.5.9",
-    "intl": "1.2.5",
-    "js-marker-clusterer": "^1.0.0",
+    "angulartics2": "^2.2.2",
+    "core-js": "^2.4.1",
+    "crds-ng2-content-block": "1.0.2",
+    "crds-reactive-auth": "crdschurch/crds-reactive-auth#development",
```
plachmann: Should we set these 2 packages to a specific version?

package.json
```diff
@@ -1,106 +1,73 @@
...
-    "awesome-typescript-loader": "3.1.3",
-    "bootstrap-sass": "3.3.7",
-    "codelyzer": "2.0.0",
-    "copy-webpack-plugin": "4.0.0",
-    "crds-ng2-content-block": "1.0.2",
-    "css-loader": "0.25.0",
-    "css-to-string-loader": "0.1.2",
-    "del-cli": "^0.2.1",
-    "dotenv-webpack": "1.4.0",
+    "@angular/cli": "^1.4.2",
```
plachmann: I love the reduced number of dev dependencies.

src/app/app.component.ts
```diff
@@ -23,15 +21,13 @@ declare var svg4everybody: any;
         <router-outlet></router-outlet>
       </div>
     </div>`,
-  styleUrls: ['../styles/application.scss'],
```
plachmann: Where did this move to?

package.json
JoeKer1: I actually had  updated crds-ng2-content-block  to the latest but that broke all our stuff because they changed the content service to return observables. I figured, since it affected so much code that should be it's own story. But you're probably right for crds-reactive-auth.

README.md
```diff
@@ -1,43 +1,111 @@
-# CrdsConnect
+# CrdsAngularBoilerplate
```
danrye: @JoeKer1 Probably want to update this

package.json
danrye: @JoeKer1 he's talking about reactive-auth and crds-styles

JoeKer1: Ahh I didn't see the styles, good catch. reactive auth isn't published to npm yet and it's using the same version as group-leader-app. We should probably get that published tho

danrye:
@JoeKer1 jerk face

PR #713, opened 2017-09-26
us9907 - leader adds new participant as member (not trial member)

plachmann:


src/app/components/pin-details/gathering/add-someone/add-someone.component.spec.ts
```diff
@@ -96,14 +96,15 @@ describe('AddSomeoneComponent', () => {
             BlandPageCause.Success,
             `gathering/${gatheringId}`
         );
+        const roleId = 16;
```
harimau777: Should this be added to the constants file?

PR #714, opened 2017-09-26
Leader and Group Resources buttons now how login and group leader guards

harimau777:
https://rally1.rallydev.com/#/66097202244d/detail/defect/155806977340


src/app/components/pin-details/gathering/leader-resources/leader-resources.component.ts
```diff
@@ -1,27 +1,20 @@
...
-
 export class LeaderResourcesComponent  {
+  constructor(private router: Router) {}
 
-    constructor() { }
-
-    public onLeaderResourcesClicked() {
-        window.location.href = LeaderResourcesUrl;
-    }
+  public onLeaderResourcesClicked() {
```
JoeKer1: Can these 2 methods be tested?

JoeKer1: or does the guard do that?

PR #715, opened 2017-09-26
US9967 - Mobile Groups Search

seancdavis:
This adds support for the expanded search view on mobile.

It's for mobile only. Desktop comes as part US9968, but merging this alone would create several regressions in connect and DDK. So we may want to let this sit here until US9968 is ready to go.


harimau777:
Have we decided if we want to use package-lock.json?  I think we had removed it since we thought that we could handle that inside of the package.json, but its not really something I've looked into.

seancdavis:
@harimau777: I would love to get rid of it—it seems to only be causing problems. I'll make that change shortly ...

PR #719, opened 2017-09-28
hide the location bar when search is kicked off

JoeKer1:


JoeKer1:
No, it is only kept track of while you're on the map right now. we'd never come back and have it expanded

PR #724, opened 2017-09-29
Defect/de4389 filters saving

JoeKer1:
This PR re-selects selected filters on initialization of each filter. So when you come back from a details page it should work better.


JoeKer1:
I agree with you. A lot of the regex stuff should probably be removed once we complete the URL params story.

PR #725, opened 2017-09-29
de4403 connect list view

plachmann:


src/app/components/list-entry/list-entry.component.ts
```diff
@@ -59,11 +59,15 @@ export class ListEntryComponent implements OnInit {
...
-      this.adjustedLeaderNames = this.getAdjustedLeaderNames(this.leaders);
-    });
-    this.proximityInfo = this.getProximityDisplayString(this.pin);
+    if (this.isSmallGroup) {
+      this.participantService.getAllLeaders(this.pin.gathering.groupId).subscribe((leaders) => {
+        this.leaders = leaders;
+        this.adjustedLeaderNames = this.getAdjustedLeaderNames(this.leaders);
+      });
+
+      this.proximityInfo = this.getProximityDisplayString(this.pin);
```
JoeKer1: should proximity info be outside the if statement?

src/app/components/list-entry/list-entry.component.ts
```diff
@@ -59,11 +59,15 @@ export class ListEntryComponent implements OnInit {
     this.isGathering = this.type === pinType.GATHERING;
     this.isSite = this.type === pinType.SITE;
     this.isSmallGroup = this.type === pinType.SMALL_GROUP;
-    this.participantService.getAllLeaders(this.pin.gathering.groupId).subscribe((leaders) => {
-      this.leaders = leaders;
-      this.adjustedLeaderNames = this.getAdjustedLeaderNames(this.leaders);
-    });
-    this.proximityInfo = this.getProximityDisplayString(this.pin);
+    if (this.isSmallGroup) {
```
JoeKer1: should this be if(this.pin.gathering) (if the pin has a group on it?) or do gatherings and small groups do different things

src/app/components/list-entry/list-entry.component.ts
JoeKer1: If so i think there's probably another if statement in there

src/app/components/list-entry/list-entry.component.ts
plachmann: The list of leaders really only applys to small groups. proximity string is also focused on small groups, so I'll change the name

PR #738, opened 2017-10-03
US10050 - Filter toggle text

zarahzachz:
Add logic to toggle the filter text label.


zarahzachz:
🤔 I'm thinking the text changing from "add filter" to "close filter" is a good enough active state, but I'll run it past Rob. Nice catch!

PR #755, opened 2017-10-11
say-hi only on person pin

plachmann:


src/app/components/pin-details/gathering/gathering.html
```diff
@@ -44,6 +44,8 @@
 
       <div *ngIf="pin.pinType !== pinType.SMALL_GROUP && !isInGathering">
         <button class="btn btn-secondary btn-block-mobile" (click)="requestToJoin()">Connect to gathering</button>
+      </div>
+      <div *ngIf="isPerson()">
         <say-hi [isGathering]="true" [isLoggedIn]="isLoggedIn" [user]="user" [pin]="pin" [buttonText]="sayHiButtonText"></say-hi>
```
JoeKer1: Is it possible to do this in the say hi component itself? We already pass in if its a gathering or not. 
Will this get rid of the Contact Host Button completely?

PR #771, opened 2017-10-14
Defect/revert back a rrow

JoeKer1:
Full disclosure: I have no idea why this is faster. But runway 2 is much faster than int with this code and does the same stuff.  Also fixed a bug where clicking on a filter took away the location bar.


src/app/components/search-bar/search-bar.component.html
```diff
@@ -76,6 +76,5 @@
       </div>
     </form>
   </div>
-</div>
-
   <app-filters [hidden]="!state.isFilterDialogOpen" [isConnectApp]="isConnectApp" (cancelFilter)="filterCancel()" (applyFilters)="onSearch(state.searchBarText)" class="connect-filters connect-container connect-layout-container"></app-filters>
+</div>
```
Doug-Shannon: Could this be it?

PR #784, opened 2017-10-23
added automation id's and fixed a build error(max to maxLenght)

simeonsena:


src/app/components/pin-details/gathering/participant-remove/participant-remove.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="fauxdal-wrapper">
   <div class="fauxdal fauxdal-fullscreen">
     <div class="container-fluid">
-      <div class="pull-right close">
+      <div class="pull-right close" data-automation-id="removeParticipantCancelXButton">
```
zarahzachz: 🤔 I don't know if these automation ids work on `div` tags. If this proves to be a problem, there's an `a` tag (line 5) that might be a better solution.

simeonsena: my old test used the CSSClass="pull-right close" and it worked. It should work as a unique cssclass.

PR #788, opened 2017-10-25
Feature/us10281 update connect switch toggle

galiotoweb:
Update connect to use new switch toggle style

Given a Connect user
When trying to see hosts in the connect tool
Then I should see a filter that uses the switch toggle from the DDK/crds-styles


zarahzachz:
@galiotoweb First, the styles look great *but* it's not functioning correctly...

So it looks like the filter is open by default (I'm not sure it's supposed to do this, but I don't know if you've had conversations with Rob on this behaviour). Once you click the filter button it hides the map/list and not the filters. Also, the filters button isn't inheriting a teal active state when selected (as it does in /groups).

nffg4rybz4

galiotoweb:
Hey Sean,  I'm happy to modify the breakpoint for mobile.

Re: the map not updating right away. That's not the current functionality either. The way the filter currently works is when you open the filter dialog tray, the tray opens you make a choice and then either click search or click apply filters. It's not activated by selecting the toggle. 

Is the functionality supposed to be changed with the new design?

galiotoweb:
Acceptance Criteria:
Given a Connect user
When trying to see hosts in the connect tool
Then I should see a filter that uses the switch toggle from the DDK/crds-styles 

Screenshots and design are in Rally story under description

On connect app, when filter button is clicked, the filter tray opens and the map is still visible. 
Selecting on or off on toggle button refreshes map with the host filter applied.

PR #791, opened 2017-10-27
added more automation ID's

simeonsena:


src/app/components/bland-page/bland-page.html
```diff
@@ -19,7 +19,7 @@
 
       <footer class="page-footer btn-group-vertical">
         <button type="button"
-                class="btn btn-secondary btn-block-mobile"
+                class="btn btn-secondary btn-block-mobile" data-automation-id="returnToResultsButton"
```
JoeKer1: This is not always return to results button. It could go to any route I think. Let me know if you still want to merge with this name.

simeonsena: I'm ok with that. As long as it's something unique. I'm not searching by button text.

PR #801, opened 2017-11-09
De4672/filter icon toggles tray

galiotoweb:
## Purpose
In /connect and /groups, the filter icon should toggle the filter tray (open and close).

Current behavior has the filter tray closing only when a filter is applied or canceled.

## Approach
New behavior: filter icon both opens and closes the filter tray on both /connect and /groups


tcmacdonald:
Closed in lieu of #803.

PR #808, opened 2017-11-20
Feature/us10899 update searchbar

zarahzachz:
Updated the styles and markup for the searchbar on Finder to use new styles. Tweak the focus states styles.

Corresponds with crdschurch/crds-styles#253


zarahzachz:
@seancdavis I believe the focus state's been fixed. It's ready for review.

PR #838, opened 2018-01-04
Add timeout back so filters will redraw map overlay

JoeKer1:
I had to add the timeout back in so that the map actually gets hidden (before it would set it back to unhidden so fast it wouldn't re-init).


src/app/components/neighbors/neighbors.component.ts
```diff
@@ -123,7 +123,11 @@ export class NeighborsComponent implements OnInit, OnDestroy {
     );
 
     this.state.setLoading(false);
-    this.state.showMap();
+
+    // This is done so that we make sure the map is hidden for filters to apply.
+    setTimeout(() => {
+      this.state.showMap();
+    }, 1)
```
Doug-Shannon: U-G-L-Y

PR #870, opened 2018-02-12
Defect/de5036 mailto error

plachmann:
## Purpose
Mailto link misbehaving for user. Can not recreate locally

## Approach
Use comma instead of semicolon


JoeKer1:
We changed this due to a different help desk ticket to be a semicolon. So it may re-break that case? Which one should we use?

PR #879, opened 2018-03-21
fix defect

JoeKer1:
actually load the congregation from the group being edited.


src/app/components/create-group/page-6/create-group-page-6.component.ts
```diff
@@ -101,6 +101,7 @@ export class CreateGroupPage6Component implements OnInit {
     this.state.setLoading(true);
     this.isSubmitted = true;
     if (form.valid) {
+      debugger;
```
plachmann: remove

PR #922, opened 2018-05-09
US13362 | Christian | css, html update DDK to match current version

christiancrawford:


src/app/components/pin-details/gathering/gathering.html
```diff
@@ -31,11 +31,11 @@
...
       </div>
 
-      <div class="btn-group push-top">
-        <input *ngIf="app.isSmallGroupApp() && isLeader" type="button" class="btn btn-secondary" value="Edit Group" (click)="onEditGroupClicked(pin.gathering.groupId)">
-        <div class="font-size-smaller push-top">
-          <a class="pointer" *ngIf="isLeader && !isInConnectApp" (click)="onEndGroupClicked()">End group</a>
-        </div>
+      <div class="push-top">
+        <button class="btn btn-white btn-outline" *ngIf="app.isSmallGroupApp() && isLeader" type="button" value="Edit Group" (click)="onEditGroupClicked(pin.gathering.groupId)">Edit Group</button>
+        <button class="btn btn-cyan-light btn-outline pointer" *ngIf="isLeader && !isInConnectApp" (click)="onEndGroupClicked()">
```
zarahzachz: `btn-cyan-light` isn't supported by the DDK. You may want to discuss the design direction here with @amplerob. I don't believe I've seen a button pattern used where we have different colored outline buttons side-by-side like this. Traditionally, we'd have a solid CTA style button next to an outline.
 
screen shot 2018-05-09 at 11 20 30 am

src/styles.scss
```diff
@@ -122,4 +122,10 @@ app-neighbors {
   background-color: transparent;
   border-color: rgba(151, 151, 151, 0.75);
   color: #979797;
-}
\ No newline at end of file
+}
+
+.btn-cyan-light{
```
zarahzachz: @christiancrawford If you end up rolling your own button style here, be mindful that you'll need to support `:hover`, `:focus` and `:active` states. In the DDK, you can see the CSS behind our button code [here](https://github.com/crdschurch/crds-styles/blob/development/assets/stylesheets/_mixins.scss#L13)

src/app/components/pin-details/gathering/end-group/end-group-confirmation.component.html
```diff
@@ -17,18 +17,14 @@
         </div>
 
         <footer class="row soft-top">
-          <div class="col-sm-6 col-sm-offset-3 btn-block-mobile text-center">
```
zarahzachz: @christiancrawford Why did you remove the `btn-block-mobile` class?

src/app/components/pin-details/gathering/end-group/end-group-confirmation.component.html
```diff
@@ -17,18 +17,14 @@
...
-                    (click)="onEndGroup()">
+            <button type="button" class="btn btn-cyan" (click)="onEndGroup()">
               End group
             </button>
 
             <!-- Cancel Button: -->
-            <button type="button"
-                    class="btn btn-outline btn-default btn-block-mobile"
-                    (click)="onClose()">
+            <button type="button" class="btn btn-outline btn-default" (click)="onClose()">
```
zarahzachz: @christiancrawford Because of the button refactor in `crds-styles`, some of the classes may not map one-for-one here. To achieve the same button as we have in INT, it looks like the class `btn-gray-light` should be used instead of `btn-default`.

src/app/components/pin-details/gathering/gathering.html
christiancrawford: reverting this change per our conversation

src/styles.scss
christiancrawford: reverting this change per our conversation

src/app/components/pin-details/gathering/end-group/end-group-confirmation.component.html
christiancrawford: because I didn't have full understand on it, reverting this change

src/app/components/pin-details/gathering/end-group/end-group-confirmation.component.html
christiancrawford: adjusted the styles here

PR #924, opened 2018-05-09
Feature/us13344

christiancrawford:


.angular-cli.json
```diff
@@ -6,7 +6,7 @@
   "apps": [
     {
       "root": "src",
-      "outDir": "dist",
+      "outDir": "../crds-maestro/apps/crossroads_interface/priv/static/js/crds_connect",
```
zarahzachz: @christiancrawford Why was this change made?

src/app/components/filters/filters.component.html
```diff
@@ -90,11 +90,9 @@ <h4 class="accordion-subheader text-uppercase font-size-smaller">Frequency</h4>
...
 
-  <div *ngIf="isConnectApp">
-    <div class="col-xs-6 soft-half-top push-quarter-top host-text">Only show hosts</div>
-    <div class="col-xs-6 host-button">
-      <app-boolean-filter [filter]="filterService.allFilters.hostOnly" [applyOnSelect]="true" (applyFilter)="apply()" [isConnectApp]="isConnectApp"></app-boolean-filter>
-    </div>
+  <div *ngIf="isConnectApp" class="form-group push-left soft-half-top">
+    <label class="control-label block">Only show hosts</label>
+    <app-boolean-filter [filter]="filterService.allFilters.hostOnly" [applyOnSelect]="true" (applyFilter)="apply()" [isConnectApp]="isConnectApp"></app-boolean-filter>
   </div>
```
zarahzachz: @christiancrawford Is this supposed to align all the way to the left like this? It feels like it could get pretty hairy on larger monitors.
screen shot 2018-05-09 at 11 56 56 am

christiancrawford: thank you for that explanation :) . . . good solution too!

PR #980, opened 2018-07-26
sorting by site name

Doug-Shannon:


src/app/components/address-form/address-form.component.ts
```diff
@@ -28,27 +29,49 @@ export class AddressFormComponent implements OnInit {
...
+      siteDetails: new FormControl(this.address.congregationId, [
+        Validators.required
+      ])
     });
 
     this.parentForm.addControl(this.groupName, this.addressFormGroup);
 
     this.crossroadsCongregationService
       .getCongregation()
+      .pipe(
```
Doug-Shannon: this pipe function is all that actually changed - everything else is linting

PR #1068, opened 2019-07-29
DE7008 - Fix env:replace

seancdavis:
Add new env:replace script that checks components source.


bin/env-replace.js
```diff
@@ -0,0 +1,45 @@
...
+
+if (env == "int" || env == "development" || env == "dev") {
+  console.log("[env:replace] No replacement for int environment.");
+  return;
+}
+
+// ---------------------------------------- | Attributes
+
+const attrOptions = {
+  files: "./dist/search-ui/index.html",
```
tcmacdonald: Is this a typo? `/search-ui`

PR #1103, opened 2019-12-09
US18506

j-brammer:
updated routing for editing and creating groups
removed the subcategory step


src/app/components/create-group/create-group.service.ts
```diff
@@ -194,25 +192,30 @@ export class CreateGroupService {
...
+  }
 
-    Object.assign(
-      this.group.attributeTypes,
-      this.group.attributeTypes,
-      jsonObject
-    );
+  public hardCodeCategories(){
+    this.group.attributeTypes = HARDCODEDSETTINGS;
+    console.log(this.group.attributeTypes);
```
j-brammer:
```suggestion
```

PR #1137, opened 2020-03-10
DE7544 updated toaster to be awesome

j-brammer:
make toast great again


src/app/services/session.service.ts
```diff
@@ -72,13 +71,6 @@ export class SessionService extends SmartCacheableService<any, number> {
               }
             })
           })
-          .pipe(
```
christiancrawford: Curious why you removed this? No opinion here just curious.

j-brammer: all that pipe was doing was basically hopping into the observable for debugging purposes. It had no real function and I realized it was console logging out a bunch of stuff we don't want to.

j-brammer: Also, catchError should bubble up properly. if this errors I don't actually have anything I need/want to do.

PR #1138, opened 2020-03-10
DE7565 | Christian | add check 'isAuthenticated' to update 'contactId'

christiancrawford:


src/app/services/session.service.ts
```diff
@@ -223,7 +223,8 @@ export class SessionService extends SmartCacheableService<any, number> {
 
   public getContactId(): number {
     let cID = +this.cookieService.get(this.contactId);
-    cID = isNaN(cID) ? this.authService.contactId : cID;
+    console.log('cID', cID);
```
janssejl: console logs yo

src/app/services/session.service.ts
```diff
@@ -223,7 +223,8 @@ export class SessionService extends SmartCacheableService<any, number> {
...
+    cID = isNaN(cID) || !this.authService.isAuthenticated ? this.authService.contactId : cID;
     return cID;
   }
```
j-brammer: lets just change `cID` to `contactId`....this ain't no C

